PR #9, opened 2017-11-28
Added admin settings

TamaraGunkel:
I added the admin settings for series name, group name, group  creation and additional ACL roles (see issues).


tobiasreischmann:
I just saw two little bugs:
- lang string aclrolesnamedesc is not defined.
- the name of the field limitvideos in the admin_form is wrong. It is initialized as uploadworkflow. This way some settings can not be saved!

tobiasreischmann:
Can you add an install.php which initially creates default ACL roles. Created should be the three default roles of the previous code.

tobiasreischmann:
I would love to have a dropdown choice for the max upload size similar to admin/settings/security.php:30-32
We need significantly higher upload sizes, than the normal maxbytes of moodle here. I think the following values could be suitable:
50MB
100MB
200MB
500MB
1GB
2GB
5GB
10GB
Unlimited

Issue #23, opened 2018-04-17
Define which ACLs should be permanent and which should be removed to manage student access


ebbertd:
Issue #22 requests to be able to manage to manage student access. To implement that it is needed to change the ACLs in Opencast. In the current block settings it is possible to define which ACLs are added to uploaded videos. This list could be extended with a checkbox to manage wether a rule is permanent (e.g. for admin or instructor access) or not (e.g. for student access).

Issue #22, opened 2018-04-17
As an instructor I need to be able to manage student access to my uploaded videos


ebbertd:
As an instructor I need to be able to manage student access to my uploaded videos. It should be possible to remove student access to a video without deleting it and it should be possible to add these rights again later on.

tobiasreischmann:
Could be solved using a visibility icon instead of the garbage bin behind each video entry. In the admin settings the ACL rules would need to extended with a flag, which tells if the ACL rules are removed for an "invisible" video.

Issue #17, opened 2018-02-05
Create series without upload


tobiasreischmann:
Offer a button under the capability of an opencast admin, to directly create the opencast series with all ACLs without uploading a video.

Issue #15, opened 2018-02-05
Setting for series identifier


tobiasreischmann:
Add a setting to the overview block, which allows to select the corresponding opencast series from a dropdown. This should only be possible with a specific capability. For a lot of series in a system, we should think about something like a searchable dropdown.

PR #27, opened 2018-04-23
New setting for series

TamaraGunkel:
I've added a setting which allows the user to select a opencast series for a course (Issue #15). If no series is defined yet, the user can create a new series (Issue #17). Changes are not applied to old videos. 
I've also added everything mentioned in Issue #22 and #23.


renderer.php
```diff
@@ -183,6 +183,27 @@ public function render_delete_acl_group_assignment_icon($courseid, $videoidentif
...
+     *
+     * @param string $videoidentifier
+     */
+    public function render_change_visibility_icon($courseid, $videoidentifier, $visible) {
+        global $USER;
+        $url = new \moodle_url('/blocks/opencast/changevisibility.php',
+            array('identifier' => $videoidentifier, 'courseid' => $courseid, 'visible' => $visible, 'sesskey' => $USER->sesskey));
+        $text = get_string('changevisibility', 'block_opencast');
+
+        if($visible) {
```
tobiasreischmann: The icon should represent the current status and not the action to be done. Compare with hide/show for course modules.

renderer.php
```diff
@@ -183,6 +183,27 @@ public function render_delete_acl_group_assignment_icon($courseid, $videoidentif
...
+    /**
+     * Render the link to change the visibility of a video.
+     *
+     * @param string $videoidentifier
+     */
+    public function render_change_visibility_icon($courseid, $videoidentifier, $visible) {
+        global $USER;
+        $url = new \moodle_url('/blocks/opencast/changevisibility.php',
+            array('identifier' => $videoidentifier, 'courseid' => $courseid, 'visible' => $visible, 'sesskey' => $USER->sesskey));
+        $text = get_string('changevisibility', 'block_opencast');
```
tobiasreischmann: Please add texts according to the action to be done, when clicking the icon. E.g:
Grant access by students
Revoke access by students

index.php
```diff
@@ -119,6 +135,9 @@
         $row[] = $renderer->render_publication_status($video->publication_status);
         $row[] = $renderer->render_processing_state_icon($video->processing_state);
 
+        $visible = $apibridge->is_event_visible($video->identifier, $courseid);
+        $row[] = $renderer->render_change_visibility_icon($courseid, $video->identifier, $visible);
```
tobiasreischmann: The icon should only be rendered, if there are non permanent ACL rules!

PR #30, opened 2018-05-22
Privacy api

TamaraGunkel:
Implements the Privacy API. The block stores temporarily data of users when a video is uploaded.


tobiasreischmann:
Since this PR is not separate from #27, it has to wait until the other one is merged. For the future, please create one branch for every pull request. Each request should only contain one feature.

Issue #31, opened 2018-05-24
Reference to repository should be deleted


tobiasreischmann:
https://github.com/unirz-tu-ilmenau/moodle-block_opencast/blob/f3afe620566a31eeb5e6c51af25014ebf5dacb3d/classes/local/event.php#L134

PR #33, opened 2018-06-08
Fixes for #31 and #37

do-you-even-curl:


do-you-even-curl:
this fixes #31

do-you-even-curl:
this also fixes #37

tobiasreischmann:
Thanks for the two fixes!

Issue #18, opened 2018-03-16
Add validation for all Configurations


tobiasreischmann:
Especially when series name is empty, Opencast will throw a 500 server error message.
Thus, we should check all configurations for validity.

TamaraGunkel:
Is there anything else that should be validated?
And should the settings also be checked before they are used?

synergy-wagner:
Hi everybody, 

I just wondered whether there is a non moodle standard admin setting page for the block. Is planned validation the reason?

Kind regards
Andi

abias:
I think #80 will solve most of these validation needs, at least if we want to check things like if a field is not empty.

PR #38, opened 2018-06-12
Added Validation

TamaraGunkel:
I've added validation of the upload workflow name, group name and series name. The validation is done when changing the admin settings.
This should cover all settings that are used by opencast and have to be valid (see. Issue #18).


TamaraGunkel:
This shouldn't have happened. I'll delete the unnecessary files.

PR #54, opened 2018-07-05
Fixes the validation of series ids at multiple stages

tobiasreischmann:
For instance editing and creating of series, but also with a warning on
the index page.


tobiasreischmann:
Fixes #54

PR #59, opened 2018-07-31
WIP : Fix the depiction of the visibility icon

NinaHerrmann:
### Previously:

-  The number of action all not permanent roles have was counted.

### Now:

- The number of roles who have all necessary actions is counted.

- It is checked whether some role has one of the actions. 

If every not permanent role has all actions the event is visible.
If no role has all actions and no role has someactions the event is hidden.
In any other case, the visibility is mixed.

Needs to be tested @tobiasreischmann .

### DIFF
--- a/classes/local/apibridge.php
+++ b/classes/local/apibridge.php
@@ -863,7 +863,7 @@ class apibridge {
     }
 
     /**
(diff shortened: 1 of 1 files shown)


tobiasreischmann:
Merged into develop

PR #61, opened 2018-08-09
Link publications to their URL unless there is also media attached.

loewis:
In the block, it would be useful if it actually showed the URL where it video is published (in particular for engage). This patch does that. When I first implemented it, it linked the "api" publication in a bogus manner. This is because the real meat of that is in the media and attachments, but there is no meaningful way to link all of them. So I decided to link nothing if there is no clear preferred link target.

### DIFF
--- a/classes/local/apibridge.php
+++ b/classes/local/apibridge.php
@@ -235,6 +235,29 @@ class apibridge {
         return $result;
     }
 
+    public function get_opencast_video_publications($identifier) {
+        $resource = '/api/events/' . $identifier . '/publications';
+        $api = new api();
+
+        $publications = $api->oc_get($resource);
+
+        $result = new \stdClass();
+        $result->publications = false;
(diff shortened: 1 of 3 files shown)


mtneug:
Sorry for the late reply.

I think the `api` publication channel links to the external API endpoint and therefore returns a JSON response. This, of course, is not useful for educators. However, this is also the case for `oai-pmh`. A particular installation might additionally change the URL or include more publication channels (e.g. `annotation-tool`) where no media is published. A link might still be helpful in this case.

We discussed this internally and came to the conclusion that a white list is probably the best general approach. The admin would configure a list of publication channels that should be displayed. Then all displayed publication channels would link to their URL.

Would you be willing to implement this feature?

tobiasreischmann:
This issue should also somehow be solved by #112.
Thus, I am closing this here.

PR #67, opened 2018-09-10
WIP: Improved how planned videos are displayed

tobiasreischmann:
This merge will allow to display videos, which are currently captured.


tobiasreischmann:
Further it displays videos, which currently need cutting. Those are identified through the following three properties:
- has_preview is true
- status is EVENTS.EVENTS.STATUS.PROCESSED
- there is only one publication, which is internal

PR #72, opened 2018-10-19
Fixed block context not being used for system blocks

t-schroeder:
You can add a block in the system context as opposed to the course
context. If you do that the upload form uses the course context. That's
arguably not correct as it doesn't respect the capabilities set in the
block context.


t-schroeder:
I'm closing this since we no longer use this solution.

PR #74, opened 2018-10-29
[WIP] Added new state "in cutting"

tobiasreischmann:
We are currently unsure, how to identify if a event is "in cutting". In our system it is currently a comment at the media package. But it is difficult to retrieve this using the external API. It would be helpful to know, if other institutes would like to have this state to be displayed, as well. And how this state is represented in their system.


tobiasreischmann:
Already fixed

PR #91, opened 2019-08-16
extended version of block opencast (New)

ferishili:
This is the latest extended version of metadata branch
It requires the major update on tools_opencast: 
https://github.com/unirz-tu-ilmenau/moodle-tool_opencast/pull/17

Also, the predefined data for the metadata catalog can be downloaded below;
[mdl_block_opencast_catalog.zip](https://github.com/unirz-tu-ilmenau/moodle-block_opencast/files/3509856/mdl_block_opencast_catalog.zip)


db/install.xml
```diff
@@ -84,5 +86,29 @@
...
       </INDEXES>
     </TABLE>
+    <TABLE NAME="block_opencast_catalog" COMMENT="this table contains the list of metadata catalog and its data types">
+      <FIELDS>
+        <FIELD NAME="id" TYPE="int" LENGTH="10" NOTNULL="true" SEQUENCE="true"/>
+        <FIELD NAME="name" TYPE="char" LENGTH="255" NOTNULL="true" SEQUENCE="false"/>
+        <FIELD NAME="datatype" TYPE="char" LENGTH="255" NOTNULL="true" SEQUENCE="false"/>
+        <FIELD NAME="required" TYPE="int" LENGTH="1" NOTNULL="true" DEFAULT="0" SEQUENCE="false"/>
+        <FIELD NAME="readonly" TYPE="int" LENGTH="1" NOTNULL="true" DEFAULT="0" SEQUENCE="false"/>
+        <FIELD NAME="default_json" TYPE="text" NOTNULL="false" SEQUENCE="false"/>
```
tobiasreischmann: You have inconsistencies in the naming of this column. In the code you often name it param_json, which leads to errors when executing the code.

db/install.xml
```diff
@@ -1,14 +1,16 @@
 <?xml version="1.0" encoding="UTF-8" ?>
-<XMLDB PATH="blocks/opencast/db" VERSION="20180704" COMMENT="XMLDB file for Moodle blocks/opencast"
+<XMLDB PATH="blocks/opencast/db" VERSION="20190718" COMMENT="XMLDB file for Moodle blocks/opencast"
```
tobiasreischmann: You need to repeat all changes within the install.xml also in the upgrade.php in order to get it to work for sites, which already have a version of block_opencast installed. Through the XMLDB Editor you can generate the code for the upgrade.php.

classes/local/upload_helper.php
```diff
@@ -76,10 +76,11 @@ public static function get_upload_jobs($courseid) {
 
         $allnamefields = get_all_user_name_fields(true, 'u');
 
-        $sql = "SELECT uj.*, f.filename, f.filesize, $allnamefields
+        $sql = "SELECT uj.*, f.filename, f.filesize, $allnamefields, md.metadata
                 FROM {block_opencast_uploadjob} uj
-                JOIN {files} f ON uj.fileid = f.id AND f.component = :component
+                JOIN {files} f ON uj.presenter_fileid = f.id OR uj.presentation_fileid = f.id AND f.component = :component
```
tobiasreischmann: Through this or you return multiple rows with the same uploadjob id. You should better JOIN to files twice. Once for the presenter and once for the presentation.

classes/local/upload_helper.php
```diff
@@ -76,10 +76,11 @@ public static function get_upload_jobs($courseid) {
...
                 JOIN {user} u ON f.userid = u.id
+                JOIN {block_opencast_metadata} md ON uj.id = md.uploadjobid 
                 AND f.filearea = :filearea AND f.filename <> '.'
```
tobiasreischmann: By resolving the upper request, please move this line also to each of the JOINS. I have no idea, why this got detached from the ON statement for the files-JOIN.

classes/local/upload_helper.php
tobiasreischmann: In future extensions (or directly if you like to) we could then improve the table for existing upload_jobs, by showing the information for both files.

tobiasreischmann:
All comments have been resolved. Closing since it is now just a duplicate of #92

PR #92, opened 2019-08-23
Extended version (separate commits)

ferishili:


classes/local/upload_helper.php
```diff
@@ -76,10 +76,14 @@ public static function get_upload_jobs($courseid) {
...
         $allnamefields = get_all_user_name_fields(true, 'u');
 
-        $sql = "SELECT uj.*, $allnamefields, md.metadata
+        $sql = "SELECT uj.*, $allnamefields, md.metadata,
+                f1.filename as presenter_filename, f1.filesize as presenter_filesize,
+                f2.filename as presentation_filename, f2.filesize as presentation_filesize
                 FROM {block_opencast_uploadjob} uj
                 JOIN {user} u ON uj.userid = u.id
-                JOIN {block_opencast_metadata} md ON uj.id = md.uploadjobid 
+                JOIN {block_opencast_metadata} md ON uj.id = md.uploadjobid
```
ferishili: https://github.com/unirz-tu-ilmenau/moodle-block_opencast/pull/91#discussion_r315761086

classes/local/form_handler.js
```diff
@@ -0,0 +1,60 @@
+require(['jquery'], function($) {
```
tobiasreischmann: For what purpose is this separate javascript needed?
Could you add some comments, please?
Also Javascript should not be placed here! Have a look at AMD modules! https://docs.moodle.org/dev/Javascript_Modules

classes/local/updatemetadata_form.php
```diff
@@ -0,0 +1,149 @@
...
+
+class updatemetadata_form extends \moodleform {
+
+    public function definition() {
+
+        $mform = $this->_form;
+
+        $metadata = $this->_customdata['metadata'];
+        
+        unset($metadata[array_search('isPartOf', array_column($metadata,'id'))]);
```
tobiasreischmann: Why is this form working so differently than the addvideo_form in terms of metadata handling? Most of the fields here have no $mform->setType call.

classes/local/upload_helper.php
tobiasreischmann: What did you want to tell me with the linked comment? I wanted to rewrote the code here to get rid of the filename and filesize being part of the uploadjob table.

updatemetadata.php
```diff
@@ -0,0 +1,93 @@
...
+$PAGE->navbar->add(get_string('pluginname', 'block_opencast'), $redirecturl);
+$PAGE->navbar->add(get_string('updatemetadata', 'block_opencast'), $baseurl);
+
+// Capability check.
+$coursecontext = context_course::instance($courseid);
+require_capability('block/opencast:addvideo', $coursecontext);
+
+$opencast = \block_opencast\local\apibridge::get_instance();
+$metadata = $opencast->get_event_metadata($identifier, '?type=dublincore/episode');
+$languages = upload_helper::get_opencast_metadata_catalog(['name' => 'language'], 'param_json');
```
tobiasreischmann: Again the question why these two metadata fields are hardcoded. Why not use the same dymanic method as for the addvideo_form?

classes/local/updatemetadata_form.php
ferishili: Updatemetadata relies completely on Opencast server metadata and gets its values from an API call. However, setType is missing and it is fixed!

updatemetadata.php
ferishili: The reason is to consider that, admin did not set the values via admin page and there are some uploaded tasks that needed to be updated in advanced! However, it tries to get its values dynamically as it defined in upload_helper::get_opencast_metadata_catalog(), in case there is nothing defined it uses the hardcoded values!

ferishili:
Hi @tobiasreischmann 
regarding to the https://github.com/unirz-tu-ilmenau/moodle-block_opencast/pull/91#discussion_r315761086, 
I am afraid I did not get your point well, as I decided to bring more file data to the existing upload_jobs table in the current extension, and show them to the user!
Cheers

classes/admin_form.php
```diff
@@ -204,6 +204,56 @@ protected function definition() {
...
+        $title = get_string('heading_readonly', 'block_opencast');
+        $mform->addElement('advcheckbox', $name, $title);
+        $mform->setDefault($name, 0);
+
+        // New catalog param json.
+        $name = 'catalogparam';
+        $title = get_string('heading_params', 'block_opencast');
+        $mform->addElement('textarea', $name, $title);
+        $mform->setType($name, PARAM_TEXT);
+        $mform->addElement('static', 'descriptionmdpj', '', get_string('descriptionmdpj', 'block_opencast'));
```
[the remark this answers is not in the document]
tobiasreischmann: I like your changes. However, most of the moodle administrators are no developers. Thus, a text such as "Parameters will be defined as options of the select element." will be of limited help. Could you add one example for every of the 4 variants? That would be great!

classes/local/form_handler.js
tobiasreischmann: Thanks, for moving this code to an AMD module. But I still do not know, why you added this javascript. The autocomplete and the filepicker modules should work through javascript from the moodle core. Usually, there is no need to add custom Javascript.

classes/local/updatemetadata_form.php
tobiasreischmann: However, even if for instance the 'language'-field is returned by the metadata API, the administrator should have the possibilities to define, if it is actually returned to the user and editable in this view. So why don't we simply render the same (a similar) form than for adding a video and fill it with default values based on the return of the metadata API?

updatemetadata.php
tobiasreischmann: > there are some uploaded tasks that needed to be updated in advanced
Do you mean at the point in time, when an admin installs this patch on their system and there are still upload_tasks to be processed? Or the case, when the admin changes the catalog in the adminsetting?

tobiasreischmann:
Regarding [#91 (comment)](https://github.com/unirz-tu-ilmenau/moodle-block_opencast/pull/91#discussion_r315761086): I think we have a good solution for this problem now. You directly included the information for both files in the rendered table of the block. I had considered this for future work. I have reverted some of your changes, since the information about filesize and the filename are always present within the mdl_files table of moodle. Thus, an extension of the SQL-query through joins was sufficient. So, we do not need to store this information twice.

classes/local/form_handler.js
ferishili: There are 2 reasons to have that javascript:
1. In case user upload a large video file via drag & drop into the filepicker, it will take sometimes to upload the file, meanwhile the Save Changes button is enable and can submit the form, therefore, our opencast team came up with the idea to disable/enable the button in this situation.
2. The autocomplete form won't take the css classes from the form-control when create it with no suggestion feature, therefore it looks ugly and makes inconsistency in UI.

updatemetadata.php
ferishili: The first scenario has been considered mainly

classes/local/form_handler.js
tobiasreischmann: ok. Thanks for clarifying. I just build the amd module and also added the minified version of it.

updatemetadata.php
tobiasreischmann: Why don't we simply process existing upload_jobs the same way, as we did before the patch? For this purpose, we only need to extend apibridge to add the filename as a title again in case the metadata of the job is empty.

tobiasreischmann: Then values for metadata fields, which are not defined by the admin, just remain empty. Before this patch, we did not set the language at all. Also the admin would have to inform the teachers to alter the metadata after the patch for existing uploads. I think this will not happen. At least not before the admin specified the metadata catalog as intended.

ferishili: I think there is a misunderstanding about the functionality of the updatemetadata, which is intended to help users edit their predefined metadata fields (in Add Video) and it is not mandatory. Additionally, there is no need to inform teachers to alter the metadata and there is no conflict with the uploaded jobs before this patch.

ferishili: However, I will remove the hardcoded values and variables and try to reform it in the way that it gets the fields from metadata catalog defined by admin and its values via the API call from Opencast.

PR #106, opened 2020-01-14
Video upload filetype check

### LINKED ISSUE #105: Upload filetype restriction not enforced
You can upload any type of file. When you upload for example a PDF as your presentation the workflow will fail and you can't remove the upload job again via the web interface. Moodle tries to upload the file and run the workflow every time the cron task is run. So you need to manually delete these upload jobs from the DB.

It's probably best to restrict the uploaded file type to videos.

t-schroeder:
Fixes #105

Also added some styling


tobiasreischmann:
Hey Tim,
What are the styling fixes for?

t-schroeder:
Hi Tobias,

without those fixes the list of supported file extensions takes up far too much space and just looks ugly. For some reason it displays one extension per line. So it would looks something like this:

```
Presenter
___________________________________
|                                 |
|      Drag & drop to upload      |
|_________________________________|
```
**Supported file types**
```
.mp4
.wmv
.flv
.webm
.avi
.ogv
.mpg
.mov
...
```

With the CSS I added the supported extensions are all listed next to each other again. So with all the changes that are in the pull request it looks essentially like it does without them, except that you have the supported file extensions listed once under each of the video upload fields, not just one overall.

t-schroeder:
Here's a screenshot of how it looks without my CSS changes:
image

t-schroeder:
And this is with the CSS changes:
image

PR #110, opened 2020-02-12
Subtitles

### LINKED ISSUE #107: Subtitles
I'm currently looking into how we can add subtitles written in the WebVTT format to Opencast videos. Both Opencast and the Paella player already support this when configured right. Here are my thoughts on how this could be implemented into the Opencast block plugin:

### Frontend
We could add another file upload field to the video upload form. Maybe in a new section "Attachments"? Since you can have different subtitles for different languages we'd need one upload field per language.

### Config
There should probably be some "dynamic" setting for the subtitle fields where you can add more fields as you like. Similar to how you configure ACLs. For each subtitle field you'd specify a) a label that will be shown in the video upload form and b) the flavor (is that the right word?) associated with the file when uploading it to Opencast as an attachment, e.g. "captions/vtt+en".

### API
Currently events are created by POSTing to /api/events but unfortunately this API endpoint doesn't support attachments. One alternative would be to use the ingest endpoint instead. Then we could upload videos like this:

1. /createMediaPackage
2. /addDCCatalog (metadata)
3. /addTrack (for each video file)
4. /addAttachment (for each subtitle file)
5. /addAttachment (ACL)
6. /ingest/&lt;workflowname&gt;

### Workflow
With the default "schedule-and-upload" workflows you need to pass extra parameters if you want Opencast to publish the subtitles to engage, e.g.:

```
"downloadSourceflavorsExist": "true",
"download-source-flavors": "captions/vtt en"
```

I haven't yet found a way to start a workflow and pass these extra parameters to it. So it's probably necessary to modify the default workflow by adding something like this in the beginning:

```
    <!-- Mark captions for publishing -->
    <operation
      id="tag"
      exception-handler-workflow="partial-error"
      description="Tagging uploaded captions for distribution">
      <configurations>
        <configuration key="source-flavors"> captions/*</configuration>
        <configuration key="target-tags">+engage-download</configuration>
      </configurations>
    </operation>
```

It seems like it should work like this. I'll be able to test this once our Opencast admin comes around to modifying the workflow.

My plan was to have distinct states for upload jobs for each step of the ingest process. (Currently there is only STATUS_CREATING_EVENT and then STATUS_UPLOADED because everything is done in one step.)

If you have any tips or suggestions please let me know.

t-schroeder:
Fixes #107


t-schroeder:
I'm closing this in favor of #139 which is the same changes just in a different fork.

PR #112, opened 2020-03-17
Link to video player from the video name in the available videos table

justusdieckmann:
I'm not sure, whether the get_player_url() function is an appropriate place to do the caching, or if I should for example move it to index.php


tobiasreischmann:
Could you add a format_text around the link?
` $row[] = format_text(html_writer::link($engageurl . '/play/' . $video->identifier, $video->title));`
This will cause the lti to be also rendered on the block page and log the teacher in.
Otherwise, a teacher will not be authorized, when clicking the link.

NinaHerrmann:
Thank you!

PR #113, opened 2020-03-18
Update list of default roles which are added to a series

abias:
This patch removes ROLE_GROUP_MOODLE_COURSE_[COURSEID] as role which is added to a series by default as it is not relevant for default installations.
On the other hand, it adds ROLE_GROUP_MH_DEFAULT_ORG_EXTERNAL_APPLICATIONS as role which is added to a series by default as this role is needed to let Moodle fetch series information.

This change is only applied to new installations.
Existing installations which are mostly configured by experienced admins in custom ways are not changed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,13 +151,14 @@ In the ACL Roles the following placeholders can be used:
  In case that multiple groups are selected in the visibility dialog, one ACL rule for every group is created.
  The basic role including the course id is removed in the case of group visibility.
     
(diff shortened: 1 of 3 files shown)


tobiasreischmann:
I also added a description, that the ROLE_GROUP_MH_DEFAULT_ORG_EXTERNAL_APPLICATIONS is a necessary role for the opencast user.

PR #120, opened 2020-03-27
Cache videodata from API Response for opencast-block

justusdieckmann:


index.php
```diff
@@ -43,6 +43,10 @@
 $coursecontext = context_course::instance($courseid);
 require_capability('block/opencast:viewunpublishedvideos', $coursecontext);
 
+// Invalidate Block cache;
+$cache = cache::make('block_opencast', 'videodata');
+$cache->delete($courseid);
```
Dagefoerde: Rather invalidate by event instead of calling `delete`: https://docs.moodle.org/dev/Cache_API_-_Quick_reference#Invalidate_keys_using_an_event

block_opencast.php
```diff
@@ -59,12 +59,27 @@ public function get_content() {
...
         $renderer = $PAGE->get_renderer('block_opencast');
-        $apibridge = \block_opencast\local\apibridge::get_instance();
-        try {
-            $videos = $apibridge->get_block_videos($COURSE->id);
-        } catch (\moodle_exception $e) {
-            $videos = new \stdClass();
-            $videos->error = $e->getmessage();
+
+        $cache = cache::make('block_opencast', 'videodata');
+        if ($cache->has($COURSE->id)) {
```
Dagefoerde: `get()` already returns `false` if the key is not set, so the double hit is unnecessary https://docs.moodle.org/dev/Cache_API_-_Quick_reference#Get_a_key

block_opencast.php
```diff
@@ -59,12 +59,27 @@ public function get_content() {
...
+            $result = $cache->get($COURSE->id);
+            if ($result->timevalid > time()) {
```
Dagefoerde: :+1:  :)

tobiasreischmann:
Merged

PR #122, opened 2020-04-06
Add LTI module feature

abias:
If enabled, teachers can add an Opencast LTI module to a course. This LTI module will be pointing to the course's Opencast series.

Note: This patch also adds some minor additions to the "Additional features" settings page and the README to better integrate itself into these files.


abias:
I am unsure why the Travis build failed for multiple environments, but these failures seem not related to this patch.

loewis:
I like the feature, but suggest to make more explicit in the labelling that this is about the Series LTI tool. Opencast also supports LTI for individual events, which, as I understand, is not covered by this change. So instead of saying "Opencast LTI Tool", I'd say "Opencast Series" or some such.

abias:
Hi @loewis ,

thank you for your feedback.

You are right, Opencast also supports LTI for individual episodes. However, I haven't seen a teacher before who worked with individual episodes and didn't want to overwhelm the teacher with just another technical term like "series".

But I will be happy to make the string changes if there is a consensus in the community or a decision by the plugin maintainers.

Thanks,
Alex

abias:
One more thing:
We have this feature in production for some days now. No technical problems happened since then.

Regarding the labeling of the feature with regards to the term "LTI", our teachers told us that they don't understand the term "LTI" at all. Thus, we are thinking about getting rid of the term "LTI" for teachers completely and just speak of "Opencast series".

I would appreciate some feedback by the plugin maintainers and will then update the language pack accordingly.

moodlebeuth:
Hi @abias: Some of our prof's are using indeed the external tool for integrating just one video or all videos in one course one by one. So I have to agree my colleague loewis. ;-)

abias:
Thank you for merging this contribution and for fixing the discussed strings afterwards in https://github.com/unirz-tu-ilmenau/moodle-block_opencast/commit/704b6e396fa40c1aba67ae005e80aa9bd758f371!

PR #129, opened 2020-04-22
Multiple UI Improvements

abias:
This PR combines multiple UI improvements which we have been running at Ulm university for 3 weeks now.

They polish the index.php (Overview page) as well as the addvideo.php (Add video page). The focus is on adding explanation texts for teachers and for unifying the look and feel of the pages.

You will find multiple commits with corresponding commit messages which describe the goal of the commit.

This combined contribution was discussed before with @tobiasreischmann and I hope that it can be merged more or less as-is.

Thanks,
Alex

PS: We have german translations ready for the added / modified strings and are ready to contribute these to AMOS as soon as this patch is merged and published in the plugin repo.

PR #130, opened 2020-04-24
Add setting to enable / disable the control of ACLs

abias:
This PR is based on PR #129 and has to be merged after #129

With this addition, the admin can configure if the teacher is allowed to control the visibility of an Opencast episode after the episode has been processed.
Additionally, the admin can configure if the episode can be restricted to course groups or not.

Note: 
This change does not have any Behat tests added.
Behat would not only have to upload a video for processing, it would also have to wait until it is really finished processing. This is slightly more complicated as I have time currently.

PS: We have german translations ready for the added / modified strings and are ready to contribute these to AMOS as soon as this patch is merged and published in the plugin repo.


abias:
I am wondering why Travis job 172.8 failed. This could be a random Travis failure and not related to this patch as the patch succeeded on the same Moodle Version / PHP Version / DB in other jobs.

tobiasreischmann:
Seems random. I restarted the build

abias:
I added another commit as I forgot to update the README

abias:
@tobiasreischmann could you please restart the single failing Travis job so that we could have a green checkmark here :)

abias:
I updated and force-pushed this PR after the review of #129.

tobiasreischmann:
Thanks for yet another nice contribution, Alex :)

PR #132, opened 2020-04-25
a video which still has the status READY_TO_UPLOAD can be deleted and removed from the cronjob queue

taraghi:
while testing/using the block opencast plugin we found out that there is no way to remove the files that are not yet uploaded to Opencast from the cronjob queue and delete them. This feature would be useful / necessary at least in following 2 scenarios:

- if some of the metadata are not set in a way that Opencast can accept them, the cronjob that transfers the uploaded videos to Opencast throws an error. AFAIK there is no way of making any changes to these waiting (and always failing) videos.
- If a teacher accidentally uploads a private video and it is still in the queue (not yet uploaded to Openncast), he or she has to wait until the video is published on Opecast. Only then he or she can delete the video, which may be too late because there may already be users who have played the video on Opecast. There is no possibility to delete the video immediately before it is uploaded to Opencast.

This patch provides the possibility on frontend of removing any files from the upload job queue and deleting them from Moodle, which still have the status STATUS_READY_TO_UPLOAD.

You can find a short discussion over that [here](https://groups.google.com/a/opencast.org/forum/#!msg/moodle-opencast/PKTFjC6CiR0/qz-0g2CHAwAJ)


ypatios:
hello,
any news on this pull request?

Thanks

tobiasreischmann:
I am afraid, I have to ask you for a bit more patience. I will not be able to have a look at any PR here for at least the next two weeks.

ypatios:
no problem – just wanted to have some feedback :)
thanks

tobiasreischmann:
Sorry for the delay.
I just looked over your PR and it looks very good so far.
Also, I think this feature is really necessary.
However, is your first use case really covered by this PR? If anything fails during the upload, the status of the job should have already been moved to a next stage. Or is every DB transaction reverted if the cronjob fails?

tobiasreischmann:
I merged the PR nevertheless, since even if it does not yet support videos that fail during the processing of a job, it still others a new and helpful feature for teachers.

PR #137, opened 2020-04-28
Translate items in Dropdown list within metadata form

Amrita1991:
This is the solution referred [here](https://github.com/unirz-tu-ilmenau/moodle-block_opencast/issues/135#issuecomment-620488264)

The dropdown list (e.g language field under Metadata in Add Video form) items are defined in install.php and hence do not get updated through the normal process of translation through lang file. My solution offers a way to remedy this. 

1. **String definitions**: Need to define string for the dropdown items in language files. I added the language_params and license_params in the english language file, so that developers can access and provide translations through AMOS. I provide the 'de' lang file as well in case people want to go by the old language folder approach.

2. **Calling the string definitions:** Next we call the translation defined in the previous step. The code doesn this by:

- Check if a field has value for param_json attribute
- if yes, it checks language files for corresponding parameters by searching in the format <_fieldname_>_params 
- if there is definition for the keyword in language file, code uses that, else it just takes the original value of the param_json attribute

### DIFF
--- a/classes/local/addvideo_form.php
+++ b/classes/local/addvideo_form.php
@@ -53,6 +53,7 @@ class addvideo_form extends \moodleform {
                     }
                 }
                 if ($field->param_json) {
+                    $field->param_json = format_string($field->param_json, true, array('filter' => 'true'));
                     $param = $field->datatype == 'static' ? $field->param_json : (array)json_decode($field->param_json);
                 }
                 if ($field->datatype == 'autocomplete') {
--- a/classes/local/updatemetadata_form.php
+++ b/classes/local/updatemetadata_form.php
@@ -42,6 +42,7 @@ class updatemetadata_form extends \moodleform {
             $param = array();
             $attributes = array();
             if ($field->param_json) {
+                $field->param_json = format_string($field->param_json, true, array('filter' => 'true'));
                 $param = $field->datatype == 'static' ? $field->param_json : (array)json_decode($field->param_json);
             }
             if ($field->datatype == 'autocomplete') {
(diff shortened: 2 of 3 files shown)


abias:
Hi @Amrita1991 ,

thank you for working on this patch.

I agree that the options for the language and the license dropdown should be translatable.
But I am wondering if your implementation is the right way to go. Especially as putting the JSON into AMOS is really risky as the translators could break the data structure.

My gut feeling would be that we should rather
* Let the admin set multilang strings for the two settings on /blocks/opencast/adminsettings.php if he needs multiple languages.
* Output the dropdown items with format_string()

Amrita1991:
> Hi @Amrita1991 ,
> 
> thank you for working on this patch.
> 
> I agree that the options for the language and the license dropdown should be translatable.
> But I am wondering if your implementation is the right way to go. Especially as putting the JSON into AMOS is really risky as the translators could break the data structure.
> 
> My gut feeling would be that we should rather
> 
> * Let the admin set multilang strings for the two settings on /blocks/opencast/adminsettings.php if he needs multiple languages.
> * Output the dropdown items with format_string()

Thank you for the response. I have implemented this according to our system. Currently in testing phase. Once it clears testing I will close the pull request as well as the issue

Amrita1991:
I implemented @abias suggestion. 
Here is a list of modifications i made based on  the suggestion

1. fetching the value with filter mode on in the forms where the language and license dropdowns are being called i.e addvideo_form and updatemetadata_form.php

2. removed the 'de' lang folder and 'language_params' and 'license_params' because the current approach doesnt need it

3. Adding hints about multilang filter in $string['descriptionmdpj'] in en lang file so that admins remember to add tags to support multilang when they add new metadata.

Apart from these, for current fields such as language and license, we have updated the values in adminsettings.php

Amrita1991:
The test failures doesnt seem to come from the changes I made

tobiasreischmann:
Hey @Amrita1991,
could you rebase your PR onto the current master?
This should resolve the issues of travis.
Thanks

ferishili:
Hi, 
please refer to #295

PR #147, opened 2020-05-19
Disables video upload

HoffmannP:
This commit introduces a new configuration setting to disable video upload.
It removes both links to uploadvideo.php (former addvideo.php) and prevents the usage of this function if disabled.

### DIFF
--- a/index.php
+++ b/index.php
@@ -137,9 +137,8 @@ if (has_capability('block/opencast:addvideo', $coursecontext)) {
     $videojobs = \block_opencast\local\upload_helper::get_upload_jobs($courseid);
     echo $renderer->render_upload_jobs($videojobs);
 
-    $addvideourl = new moodle_url('/blocks/opencast/addvideo.php', array('courseid' => $courseid));
-    $addvideobutton = $OUTPUT->single_button($addvideourl, get_string('addvideo', 'block_opencast'));
-    echo html_writer::div($addvideobutton);
+        $addvideourl = new moodle_url('/blocks/opencast/uploadvideo.php', array('courseid' => $courseid));
+        $addvideobutton = $OUTPUT->single_button($addvideourl, get_string('addvideo', 'block_opencast'));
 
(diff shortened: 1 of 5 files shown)


HoffmannP:
@Dagefoerde can I restart the tests as well? Looks to me as if I have the same problem as @aliyevTU.

Dagefoerde:
The problem is the same, but I may have been overconfident as some jobs are failing.  Please wait a bit ;)

HoffmannP:
No problem

Dagefoerde:
It looks better now. Feel free to rebase on top of the current master or to merge the current master into your branch - either way is fine.
Thanks for your work!

HoffmannP:
After all checks have passed (thank you @Dagefoerde ), is there anything I could do to support merging any further? Is complete code coverage mandatory?

tobiasreischmann:
Thanks for your contribution.
However, I don't see the direct advantage of this patch, since both uploads can already be restricted using Moodle capabilities.
Since the plugin already has so many settings, I would like to do not increase this list unnecessarily.
If you want, you could ask in the google group, whether anyone else wants to see this feature integrated, which could convince me to merge it anyways.

BerengarWLehr:
How can you disable the video-upload with Moodle capabilities?

tobiasreischmann:
For each role in you Moodle system you can set, which capabilities this role has and with that, what a user is allowed to do if the role is assigned to the user in a given context.
The relevant capability is block/opencast:addvideo. If you remove this capability from any role in you system, no one (except the admins) are allowed to upload videos.

tobiasreischmann:
See YOURMOODLE/admin/roles/manage.php

tobiasreischmann:
@HoffmannP Did you manage do deactivate the upload using the capabilities?
Do you think that this PR is still needed?

HoffmannP:
I think we don't need this feature anymore so I won't test any further

PR #155, opened 2020-09-24
Relabel the feature to delete the video before the transfer to Opencast

abias:
This commit changes a string for a label and heading and button which said "Delete video permanently".
As this feature is used to delete videos before the transfer to Opencast, the substring "permanently" is not really accurate and will be improved with this change to "Delete video before transfer to Opencast".


abias:
Well, I have to close this PR as I detected some side effect onto the feature of deleting processed videos.

PR #206, opened 2021-06-23
Improve understandability of the 'Video already uploaded' widget

abias:
This patch improves the existing 'Video already uploaded' widget which is shown in addition to the chunk upload widget in these aspects:

1. Both widgets' positions are switched, the checkbox is now shown below the upload widgets. This is done to focus the user on the upload task, not on the possibility that he can switch the widgets.
2. The checkbox gets an admin setting which lets the admin enable / disable the checkbox. Not all admins will necessarily want to provide the Moodle file picker as alternative.
3. The label of the checkbox is changed to 'Use Moodle default file picker to access all repositories'. The original string 'Video already uploaded' was not wrong, but it was limited to only one single use case: The possibility to access other files which already existing within Moodle. The new label should make it clearer what to expect when the checkbox is checked. Anyway, the admin can always change the label within the local language customization.

Furthermore, this patch adds a hideif rule to the block_opencast/uploadfilelimit setting to only show it if the chunk upload feature is enabled.


abias:
I am aware that quite all automated checks failed here, but these tests already fail on the current master branch. Thus I do not expect this PR to be the one to blame.

TamaraGunkel:
Hey,
thanks for the PR.
I like the improvements!

abias:
Thanks for merging this PR so quickly and for spotting the surplus bracket in the language pack, @TamaraGunkel !

Issue #151, opened 2020-07-30
Feature request: imported series


loewis:
It would be desirable that importing the block from another course would only change the ACLs on the series, rather than creating a new series and copying all videos. The advantage would be two-fold

- storage requirements decrease

- links to the video imported from the other course (either through the repository, or an LTI, or direct linking) would remain accessible after being imported into the new course along with the block

I propose that this is offered as an option on import, as an alternative to copying.

I then propose that the feature provided by introducing "imported series" into the block_opencast. Each block instance would have one "primary" series (as it does now), plus arbitrary many "imported" series. Adding and deleting will only be possible for videos in the primary series; Videos in imported series can only be turned visible/invisible (by changing the ACL).

As the database structure, there will be a table of course IDs that have been imported into a block. In the code, it will be possible to determine all series that have been imported, but also all courses that a series has been exported to. When a video is added to a series, its ACL reflects also the courses that the series has been exported to, giving true linking (in the sense that all modificiations to the source reflect in all importers).

In the UI, there are several alternatives for presenting the import. Clearly, the delete button is gone. Instead, it could be possible to add a link icon linking to the source block, or the source course where the video is imported from.

I'd be willing to implement this feature when I get an indication that such a change is agreeable.

abias:
Just as a quick note: I have just created a PR on #161 which realizes an import wizard.
I am aware that #161 does not fully solve this issue here, but it is worth mentioning anyway.

loewis:
In a discussion today, a different implementation strategy was proposed. The import of the opencast block from an existing course shall create a block with the **same** series ID as the source course. In order to make a video visible, the ACL must be edited to make the video available to the new course, while not affecting ACEs for other courses. Additional requirements identified:

- the feature must be optional, on a system-wide level, whether this kind of import shall be used.
- it would be desirable to see, in the block, which other courses share the same series.

ferishili:
Hi @TamaraGunkel, could you please assign this issue to me?

ferishili:
Thank you. I would like to also ask if this feature (https://github.com/Opencast-Moodle/moodle-tool_opencast/commit/61d15eed206e94c5f24074e69d57529fe3e72bdd) works fine and I can use it for this feature?

TamaraGunkel:
Well, it's in development. It does already work to a certain extend but you can expect bugs (however mostly in the block not the tool itself).
I think its a design decision on your side. My plan is to include the series feature in the next release so you can use it.
[...]

PR #212, opened 2021-06-24
New Import approach with ACL change, Solves #151

ferishili:


ferishili:
This PR is not yet suitable to be merged into master, therefore I have created another PR #213 which is targeting the branch issue-169. We close this!

PR #213, opened 2021-06-25
new import feature by changing series and events ACL #151

ferishili:


settings.php
```diff
@@ -700,6 +700,48 @@
...
+        $importvideossettings->add($select);
+
+        if ($CFG->branch >= 37) { // The hide_if functionality for admin settings is not available before Moodle 3.7.
+            $importvideossettings->hide_if('block_opencast/importmode',
+                'block_opencast/importvideosenabled', 'notchecked');
+        }
+
+        // Import Videos: Enable manual import videos feature in ACL change mode.
+        $importvideossettings->add(
+            new admin_setting_configcheckbox('block_opencast/importvideosmanualenabledwithacl',
```
TamaraGunkel, line 735: Why did you create an additional setting for the manual ACL import instead of doing it like the core import where you have one setting for both import modes?
I think it's easier for admins to decide on the Import mode first and then select where the import should be enabled.
My suggestions:

- [ ] Position Import Mode select element at the top of the settings below the general 'importvideosenabled' setting
- [ ] Use the setting 'importvideosmanualenabled' also for the ACL mode (requires changes in importvideosmanager and other places as well)

lang/en/block_opencast.php
```diff
@@ -263,17 +283,26 @@
...
 $string['importvideos_settingenabled'] = 'Allow video import';
 $string['importvideos_settingenabled_desc'] = 'If enabled, teachers can import existing video files from other Moodle courses to their Moodle course. You have to further set a duplication workflow and enable one or both of the two following settings to manually import videos or to import videos within the Moodle core course import wizard. If you don\'t enable any of these, this setting will have no effect.';
```
TamaraGunkel, line 307: Thats still tailored to the duplication mode although it is used for both. Please use a more generic text for 'importvideos_settingenabled_desc'.

lang/en/block_opencast.php
```diff
@@ -263,17 +283,26 @@
...
 $string['importvideos_importepisodecleanupfailed'] = 'At least one of the existing Opencast episode modules within this course could not be cleaned up. Please try again or contact your Moodle administrator.';
 $string['importvideos_importseriescleanupfailed'] = 'At least one of the existing Opencast series modules within this course could not be cleaned up. Please try again or contact your Moodle administrator.';
 $string['importvideos_processingexplanation'] = 'These video files will be duplicated in Opencast and then made available in this course.';
 $string['importvideos_sectionexplanation'] = 'In this section, you can import existing video files from other Moodle courses to this Moodle course.';
+$string['importvideos_aclprocessingexplanation'] = 'The seriesid of the selected course will be used in this course as secondary seriesid and the ACLs of its videos will be changed accordingly.';
+$string['importvideos_aclsectionexplanation'] = 'In this section, you can import existing video files from other Moodle courses to this Moodle course.';
```
TamaraGunkel, line 303: That's exactly the same string as 'importvideos_sectionexplanation'. Please delete it and use the existing one.

index.php
```diff
@@ -510,9 +510,19 @@
...
         echo $OUTPUT->heading(get_string('importvideos_importheading', 'block_opencast'));
 
-        // Show explanation.
-        echo html_writer::tag('p', get_string('importvideos_sectionexplanation', 'block_opencast') . '<br />' .
+        // Get import mode from the admin setting.
+        $importmode = get_config('block_opencast', 'importmode');
+        // Provide different explanation based on import mode.
+        if ($importmode == 'acl') {
+            // Show explanation for ACL Change approach.
+            echo html_writer::tag('p', get_string('importvideos_aclsectionexplanation', 'block_opencast') . '<br />' .
```
TamaraGunkel, line 518: When using the same string (see below) 'importvideos_sectionexplanation', you can move this line before the 'if' so that we have less duplicated code (even if it's only one line).

TamaraGunkel (changes requested):
Hey,
I started reviewing the new settings. Please have a look at my comments.
I'll continue to test and review the remaining parts in the following days.

classes/local/apibridge.php
```diff
@@ -1791,4 +1849,226 @@ public function get_duplicated_episodeid($workflowid) {
...
+        return true;
+    }
+
+    /**
+     * Retrieves the series data by series identifier.
+     *
+     * @param int $seriesid
+     * @return null|string id of the series id if it exists in the opencast system.
+     */
+    public function get_series_data_by_identifier($seriesid) {
```
TamaraGunkel, line 2058: Is there any reason why don't you use `get_series_by_identifier`? It's doing exactly the same except that it doesn't check the returned http code. However, you could have added this to the old method.
Please use the existing function in order to avoid duplicated code, if possible.

classes/local/importvideosmanager.php
```diff
@@ -419,4 +456,137 @@ public static function duplicate_videos($sourcecourseid, $targetcourseid, $cours
...
     }
+
+    /**
+     * Helperfunction to get the series object and process it with rendered later on in the summary.
+     *
+     * @param int $sourcecourseid The source course id.
+     *
+     * @return object
+     */
+    public static function get_import_source_course_series_object($sourcecourseid) {
```
TamaraGunkel, line 467: What is the use case of this function?
It seems as it does exactly like `get_course_series` from `apibridge`. Please try to use existing code as much as possible.

classes/local/apibridge.php
```diff
@@ -1791,4 +1849,226 @@ public function get_duplicated_episodeid($workflowid) {
...
+            return false;
+        }
+
+        // Get the current defined roles.
+        $roles = $this->getroles();
+        foreach ($roles as $role) {
+            // Initialize the role object to have a better grip.
+            $roleobject = new \stdClass();
+
+            if (strpos($role->rolename, '[USERNAME]') !== false ||
```
TamaraGunkel, line 1945: I don't think you need this if statement. Using the code within the if block should give you the same result for both branches.

classes/local/apibridge.php
```diff
@@ -1791,4 +1849,226 @@ public function get_duplicated_episodeid($workflowid) {
...
+        try {
+            // Get the current record.
+            $mapping = seriesmapping::get_record(array('courseid' => $courseid, 'series' => $seriesid, 'isdefault' => '0'));
+
+            // If the mapping record does not exists, we create one.
+            if (!$mapping) {
+                $mapping = new seriesmapping();
+                $mapping->set('courseid', $courseid);
+                $mapping->set('series', $seriesid);
+                $mapping->set('isdefault', 0); // It is not the default series of this course.
```
TamaraGunkel, line 2040: I think it would be good if the series is set as default series if we do not have a (default) series in the target course yet. It should not be possible to have a course without a default series if there are series associated with this course. So, if the target course was just created (=no default series exists) and the teacher imports the series, the series should be set as default.

classes/local/apibridge.php
```diff
@@ -1791,4 +1849,226 @@ public function get_duplicated_episodeid($workflowid) {
...
+    /**
+     * Map the seriesid to the course that has been imported to, by assinging the series as secondary.
+     *
+     * @return int $courseid id of the course being imported to.
+     * @param string $seriesid series id.
+     */
+    private function map_imported_series_to_course($courseid, $seriesid) {
+        try {
+            // Get the current record.
+            $mapping = seriesmapping::get_record(array('courseid' => $courseid, 'series' => $seriesid, 'isdefault' => '0'));
```
TamaraGunkel, line 2033: It would be better if you're using `seriesmapping::get_record(array('courseid' => $courseid, 'series' => $seriesid), true)`. This allows you to find any mapping for the course/series independent of the default status. That's important as we don't want to have the same series twice in a course.

The `true` must be added because things are a bit complicated with the multiple series features because I try to guarantee backwards compatibility. The `true` parameters allows you to query the database without setting a filter on `isdefault`.

ferishili:
> Hey,
> I started reviewing the new settings. Please have a look at my comments.
> I'll continue to test and review the remaining parts in the following days.

Hi, 
Thanks for the reviews, I am starting the corrections right away.

TamaraGunkel:
Great, thanks. The other stuff is looking good!

classes/local/importvideosmanager.php
ferishili: In order to maintain consistency in `classes/local/importvideos_step2_form_acl.php:62`, I decided to use the function `importvideosmanager::get_import_source_course_series_object` in which I get the apibridge instance and with new changes I only call the `get_course_series` method only.
As you can see there is no apibridge instance inside `importvideos_step2_form_acl` or any other importvideos steps. However, if you see no point of doing that we can change it and bring it in.

TamaraGunkel: I'd prefer to change it as it makes the code more complicated. It's totally fine to use the apibridge in the form even if its not used during the other import steps.

classes/local/importvideosmanager.php
```diff
@@ -141,15 +147,39 @@ public static function is_enabled_and_working_for_coreimport() {
...
+    
+            // If the workflow does not exist, then the feature is not working.
+            if (!$workflowexists) {
+                // Inform the caller.
+                return false;
+            }
+        } else if ($importmode == 'acl') {
+            // If ACL Change is selected as the import mode.
+
+            // TODO: Add necessary validations for ACL here...
```
TamaraGunkel, line 180: What about the Todo? Is there missing something here?

ferishili: Currently, there is nothing there, but I wrote it for future implementations. We can remove it if not wanted!

TamaraGunkel:
It's fine now. Thanks for the contribution!

ferishili:
Hi @TamaraGunkel 
thanks for the review and everything.

Issue #214, opened 2021-07-07
Predefined Presenter metadata with name of current user in add video page


ferishili:

ferishili:
It is somehow a duplicated issue (https://github.com/Opencast-Moodle/moodle-block_opencast/issues/184) and should be followed in that issue!

Issue #184, opened 2021-05-07
Support real autocompletion in the creator and contributor metadata fields


abias:
As a teacher, when I upload a video, I have the possibility to set the creator and contributor metadata fields. These are (by default) provided as autocompletion fields. However, there isn't any autocompletion, I do never get any proposed values. Currently, the autocompletion field type is just (ab)used as a widget to add multiple separated strings.

I would appreciate if the field could either provide some autocompletion like a) the creators and contributors which I have already used for other videos in my course or b) the user fullnames of the teachers in my course or c) the creators and contributors which already exist in the Opencast backend.

If this is not possible or desired, the field type should be changed to something else which does not provoke false expectations.

ferishili:
Hi @TamaraGunkel,
I would like to work on this issue. Could you please assign this issue to me?
Thanks in advance.

PR #215, opened 2021-07-07
predefined creator metadata, solves #214

ferishili:


ferishili:
It is has to be followed in https://github.com/Opencast-Moodle/moodle-block_opencast/issues/184.

PR #217, opened 2021-07-08
Autocomplete suggestions for creator and contributor metadata fields in add video page, solves #184

ferishili:
#184

### DIFF
--- a/classes/local/autocomplete_suggestion_helper.php
+++ b/classes/local/autocomplete_suggestion_helper.php
@@ -66,7 +66,7 @@ class autocomplete_suggestion_helper
         global $COURSE, $DB;
 
         // Initialize the array list to return.
-        $suggestionlist = [""];
+        $suggestionlist = [];
 
         // Get the role of teachers.
         $role = $DB->get_record('role', array('shortname' => 'editingteacher'));
@@ -167,8 +167,8 @@ class autocomplete_suggestion_helper
             }
         }
 
-        // Finally, we remove the duplicated values from the return list and return it.
-        return array_unique($suggestionlist);
+        // Finally, we return the suggestion list.
+        return $suggestionlist;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


ferishili:
Hi @TamaraGunkel,
Please take a look and give me your feedback for this PR, thanks.

classes/local/autocomplete_suggestion_helper.php
```diff
@@ -0,0 +1,211 @@
...
+    /**
+     * Get the fullname suggestions of the teachers in the course.
+     *
+     * @return array
+     */
+    private static function get_suggestions_from_course_teachers() {
+        global $COURSE, $DB;
+
+        // Initialize the array list to return.
+        $suggestionlist = [""];
```
TamaraGunkel, line 69: Simply use []. I think that's cleaner.

classes/local/autocomplete_suggestion_helper.php
```diff
@@ -0,0 +1,211 @@
...
+                // Extract the contributor object from filtered array.
+                $contributormetadataobject = array_values($contributorarray)[0];
+
+                // Merge the values of the object with the returned list.
+                $suggestionlist = array_merge($suggestionlist, $contributormetadataobject->value);
+            }
+        }
+
+        // Finally, we remove the duplicated values from the return list and return it.
+        return array_unique($suggestionlist);
```
TamaraGunkel, line 171: You don't need `array_unique` here because you are calling it in `get_suggestions_for_creator_and_contributer` anyway. It's not a big deal but if we can optimize the code and make it faster, it's always a good idea.

classes/local/autocomplete_suggestion_helper.php
```diff
@@ -0,0 +1,211 @@
...
+            }
+
+            // If the video object has contributors, we merge them to the return array list.
+            if (!empty($video->contributor)) {
+                $suggestionlist = array_merge($suggestionlist, $video->contributor);
+            }
+        }
+
+        // Finally, we remove the duplicated values from the return list and return it.
+        return array_unique($suggestionlist);
```
TamaraGunkel, line 209: Same as above

TamaraGunkel (changes requested):
Hey,
sorry, I had a lot to do during the last week.
However, the PR looks good. Thank you. I just added some minor remarks.

PR #218, opened 2021-07-14
Event process status notification after upload, fixes #111

### LINKED ISSUE #111: Notification after video has been processed
One of our users suggested to add notifications for when a video upload/processing finishes. Moodle's built-in notification system should probably be used for this. It would also be necessary to regularly poll the workflow status for uploaded videos via a cronjob. I'm not sure who should get the notification: only the uploader or each teacher in the course? Either way, I think this would be a good addition to this plugin.
What do you think about a feature like that?

P.S.: Maybe there could also be another type of notification that gets sent when another teacher uploads a video in one of your courses (or when a video is uploaded to the Opencast series, but not through Moodle)?

ferishili:
https://github.com/Opencast-Moodle/moodle-block_opencast/issues/111

### DIFF
--- a/classes/local/eventstatus_notification_helper.php
+++ b/classes/local/eventstatus_notification_helper.php
@@ -1,0 +1,226 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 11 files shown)


ferishili:
Hi @TamaraGunkel,
I need your feedback on this PR as well, thanks

PR #219, opened 2021-07-20
Add table legend with modal feature to improve ux in overview page, fixes #199

### LINKED ISSUE #199: Improve Opencast overview page UX by replacing icons with Bootstrap badges
Currently, the Opencast overview page uses icons to inform the teacher about the processing status and visibility status of the videos. The icons have tooltips, but they still are cryptical to some teachers at first:

grafik

I would propose to re-think this approach and use Bootstrap badges (https://getbootstrap.com/docs/4.0/components/badge/) instead of icons. Teachers are especially used to see badges on the course overview pages for visibility information and we would be able to ship some more and clear information to teacher by using short texts within the badges.

Here's a quick mockup how this could look like:
grafik

ferishili:
#199


TamaraGunkel:
Hey,
thanks for the PR. I have multiple recommendations:
1. I don't think we need an admin settings (we already have too many). As there are always going to be new teachers who don't know the plugin and old teachers who don't need an explanation, the admin is not in the position to decide if a legend is useful or not. Thus, we should always provide the legend.
2. Personally, I think the Capturing icon is annoying. I didn't know that we had a blinking icon when suggesting a legend below the table. Sorry for that. But I do like the modal legend. Thus, I would suggest that we only keep the legend in the modal. You could add a help button (as in the Metadata Setting table) behind the `Status` and `Visibility` column name. Clicking the icon does not open a popup like in the metadata setting table but the modal you already have implemented.
3. Please recheck the explanations for the status. Some explanations like `Processing done` or `Processing failed` don't provide more information than their label `Succeeded` or `Failed`. If teachers need an explanations, it should be a little bit longer (one sentence at least). For example, you could add the hint for the failed status, that teachers can get into contact with the support to resolve this.
4. And just list minor remark: If you're using full sentences as in the visibility explanations, please end them with a dot.

ferishili:
Hi @TamaraGunkel,
Please take a look at the new corrections. BTW, thanks for your time, reviews and the PR you have merged so far.

TamaraGunkel:
Thanks for the improvements. I like it a lot!

PR #220, opened 2021-07-21
casting maxbytes to int for chunkupload elements

t-schroeder:
Since https://github.com/Opencast-Moodle/moodle-local_chunkupload/commit/a955d12b44457ea934df67d8ad2e2863a66a6329 local_chunkupload now requires the maxbytes passed to it to be an integer. So we just cast it to an integer after fetching it from the DB (via get_config) where it's stored as a string.

### DIFF
--- a/classes/local/addvideo_form.php
+++ b/classes/local/addvideo_form.php
@@ -144,7 +144,7 @@ class addvideo_form extends \moodleform
             }
         }
 
-        $maxuploadsize = get_config('block_opencast', 'uploadfilelimit');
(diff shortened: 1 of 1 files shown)


TamaraGunkel:
Hey,
that's indeed an issue. Thanks for creating the PR. It'll be included in the next release. However, this will take a few weeks since I'm currently working on some major changes.

t-schroeder:
Thanks for merging this. We'll apply this change independent of the plugin release anyway, so that's nto a problem.

PR #221, opened 2021-08-19
Fix hide_if regression in settings.php

abias:
This patch fixes a regression with the hide_if feature in settings.php which broke when the multiple-instance feature was integrated.


TamaraGunkel:
Hey,
thanks for making me aware of that. I missed it when renaming the settings.
I've fixed it in 34ec95aeed4451ec91baf2f1f9f226d59c9a4d78 since your PR is missing the `_` in the setting names.

PR #234, opened 2021-10-29
OC Editor integration

### LINKED ISSUE #98: Feature: Integrate video editor
*This idea was raised at Opencast DACH 2019*

A clean solution would require extensive work in Opencast/on an external video editor, so this is a long term feature request.

ferishili:
This patch integrates the stand alone OC-Editor into the moodle-block_opencast plugin via LTI and redirect.
This fixes #98 

A new section "Opencast Editor integration" is added to the "Additional Feature" config setting.
A new action button is added into the action dropdown menu inside opencast video overview, called "Video Editor" right after "Update metadata".
There are new language strings related to the changes.

**To use this patch, you would need to...**

-  use the latest version of moodle-tool_opencast, which support multiple OC instances.
-  set LTI credentials for the Editor Integration in setting.
-  make sure event (Video) is successfully processed.
-  make sure event (Video) has internal publication.
-  make sure new language strings are applied.

### DIFF
--- a/classes/local/apibridge.php
+++ b/classes/local/apibridge.php
@@ -1656,6 +1656,25 @@ class apibridge
     }
 
     /**
+     * The allowance of using the opencast editor.
+     * @param object $video Opencast video
+     * @param int $courseid Course id
+     * @return bool the capability of updating!
+     */
+    public function can_edit_event_in_editor($video, $courseid) {
+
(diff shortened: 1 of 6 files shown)


TamaraGunkel:
Thanks for the PR @ferishili  and thanks for the review @mwuttke!

PR #236, opened 2021-11-18
Editor wo lti

TamaraGunkel:


lkiesow:
Hi @TamaraGunkel, can you add a short explanation what this pull request does?

TamaraGunkel:
Sure, sorry for not providing one in the first place.

Since we're not using LTI for authentication, the PR simply allows the redirection to the editor without LTI authentication. LTI is still used if the LTI settings (consumerkey + secret) are set.

lkiesow:
Thanks for the description

PR #240, opened 2021-12-02
Access video static file from video list

### LINKED ISSUE #58: Video URLs in overview page
Hello,
we are wondering if you could add a column to the list of released videos at `blocks/opencast/index.php` that contains the URL to each video.
The reason we would like to have such a feature is that we don't trust our users to find the video they want to embed (in the Atto editor) via "Media", "Browse repositories..." and "Opencast".

Is this something you would consider?

If not, would you instead consider refactoring the way this table is generated in `index.php`: https://github.com/unirz-tu-ilmenau/moodle-block_opencast/blob/f5328cd42241ccf68f324873bf9c154ac9a2af84/index.php#L157-L194 so that the table is created by a method of the renderer and/or a template? This way we would be able to override the renderer or template in our theme and add the desired column this way.

ferishili:
This PR fixes #58 and provides the feature to access the video static file in video list in a form of a button from within the video list. 
(NOT TESTED)
**Settings:** 
- A new section is introduced in admin settings under Additional features called "Access Video static file"
- Admin can enable/disable this feature.
- LTI Authentication settings are provided, in case the **Secure Static Files** setting is enabled.

**Usage**
- When the configuration and requirements are met, a button appears in the action column inside the Video list.
- The button redirects to a new page, in order to check the requirements and prepare the call to access the video static file.

**New functions**
- A new function to get and consume services of the opencast instance is introduced in apibridge class.
- A new function to create a customized tool_opencast api class is introduced in apibridge class.

**To improve**
- It might be useful to have a central config place to store LTI configs in order to use for different settings such as (Access video static file, Stand-alone Editor & Studio)

### DIFF
--- a/accessvideo.php
+++ b/accessvideo.php
@@ -60,7 +60,10 @@ $video = $apibridge->get_opencast_video($identifier);
 
 // Check if the access to the video static file is possible.
 if (!$apibridge->can_access_video_static_file($video->video, $courseid)) {
-    redirect($redirecturl, get_string('staticvideofileinvalidconfig', 'block_opencast'), null, \core\output\notification::NOTIFY_ERROR);
+    redirect($redirecturl,
+        get_string('staticvideofileinvalidconfig', 'block_opencast'),
+        null,
+        \core\output\notification::NOTIFY_ERROR);
 }
 
 // Get the search service from the opencast instance.
@@ -68,7 +71,10 @@ $searchservice = $apibridge->get_services('search');
 
 // Check if the search service is available.
 if (!$searchservice) {
-    redirect($redirecturl, get_string('staticvideofilenosearchservice', 'block_opencast'), null, \core\output\notification::NOTIFY_ERROR);
+    redirect($redirecturl,
+        get_string('staticvideofilenosearchservice', 'block_opencast'),
+        null,
+        \core\output\notification::NOTIFY_ERROR);
 }
 
 // Create custom configs to get a custom instance from tool_opencast\local\api.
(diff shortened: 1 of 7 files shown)


ferishili:
NOTE: This PR is not yet tested and finalized, there might be a few changes for this feature including the range of users that can access the video files and so on.

TamaraGunkel:
Hey,
is that PR still WIP?

ferishili:
Hi,
unfortunately yes, I am still waiting for approval. However, we could also discuss and finalize the PR.

TamaraGunkel:
Okay. thanks. I just wanted to know if I should already review it. If you don't mind, I'd wait until you got the approval before reviewing because there are other things I want to finish first.

ferishili:
That is totally fine, I will update you as soon as it is approved, thanks a lot.

TamaraGunkel:
Hey @ferishili,
any news on this PR?

ferishili:
HI @TamaraGunkel,
I am waiting for the other-party to conduct the testing, unfortunately their Opencast test-instance is not ready yet.
Let me know if it is needed/changes are required for this PR.

TamaraGunkel:
Hey,
I have some questions regarding this PR. Personally, I understood the issues rather in the way that the link to the engage player should be shown and not to the static file. I already used a part of your code in the newest release to realize this.
However, I don't see the use case for links to the static files yet. What is the advantage over the already existing download feature?
Apart from that, is it really necessary to use the internal Opencast API? If possible, it would be better if the plugins are only dependent on the external API.

ferishili:
> Hey, I have some questions regarding this PR. Personally, I understood the issues rather in the way that the link to the engage player should be shown and not to the static file. I already used a part of your code in the newest release to realize this. However, I don't see the use case for links to the static files yet. What is the advantage over the already existing download feature? Apart from that, is it really necessary to use the internal Opencast API? If possible, it would be better if the plugins are only dependent on the external API.

Hey, if I recall it correctly, the use case for this was not only playing the video from the internal API but also to use that link in other places for example in H5P etc. I created a new local plugin to satisfy this matter but it is still pending for approval in: https://tracker.moodle.org/browse/CONTRIB-8763
I am still waiting for the other party who requested it to conduct the testing, but unfortunately, it appears that there is not a proper test instance for them to do that.
I would suggest to wait for this PR and the other party to explain more detailed use case before closing it.

TamaraGunkel:
The issue with the static files is that they are not necessarily as static as it appears to be. The URLs might change when workflows are started on the event. Thus, we (from educast) are currently trying to get a more standardized PR merged into core H5P.
However, if this is the way the other party wants to go, we can discuss about integrating it. As you said, it's reasonable to wait until they can explain the use case in more detail.

ferishili:
This PR is outdated and a newer approache is introduced in #298.
Please refer to the new PR for this feature.

PR #241, opened 2021-12-08
start date to be edited, fixes #182

### LINKED ISSUE #182: Allow start date to be edited after the video upload
As a teacher, I am able to set the start date of a video during the initial upload of the video on addvideo.php.
Additionally, when the video is processed and published, I am able to edit the metadata of the video by clicking on the cog icon of the video.

However, I am not able to edit the start date of the video here.
If I picked the wrong date during the initial upload and need to change it, I have to delete the video and upload it again.

ferishili:
This PR provides the capability to edit the start date of an event in update form.
**How does it work?**
- First it extracts the startdate and starttime from the metadata passed to the updatemetadata_form.
- Knowing that the start date and time are in UTC, we convert it to the user timezone, in order to present it correctly.
- Then we show the Start Date element in update metadata form with timezone applied.
- After saving the changes, start date will be again converted to UTC timezone in order to pass the correct metadata to Opencast by performing `update_event_metadata` function.


TamaraGunkel:
Thank you for contributing!

PR #242, opened 2021-12-10
pass return params to Studio, fixes #152

### LINKED ISSUE #152: Pass return URL to Studio (and maybe not open Studio in a new tab anymore)
A couple of weeks ago, I added a feature to Studio that makes it possible to pass a "return URL" to Studio (https://github.com/elan-ev/opencast-studio/pull/674). Studio then renders an additional button "Exit and go back" after up- or downloading the recording. This avoids the user getting trapped in a UI dead end. 

It would be great if this Plugin could support this feature, as I guess it would improve usability. To be honest, I never used Moodle, but quickly searching through the code, it looks like this Plugin simply opens Studio in a new tab (`target="_blank"`) and passes only the `seriesId`. To pass the return URL, you simply add two new GET parameters:

- `return.target`: the URL-encoded URL that the button in Studio links to.
- `return.label`: a short description where the return link leads. This is optional, but can be used to improve the label of the button in Studio. The label is as follows:
  - With `return.label=Moodle`: "Exit and go back to Moodle"
  - Without return label: "Exit and go back to previous page"

Note that the label is slightly different in different languages. If you have information about the user's language, you can of course also pass something like "Moodle course" as label.

Full example (the return URL is a guess, I have no idea how the real URL would look like :smiley:):

```
https://opencast/studio/
   ?upload.seriesId=c6e9b5ab-a9eb-4e6b-9854-01b6886e0587
   &return.label=Moodle
   &return.target=https%3A%2F%2Fmoodle.example.de%2Fplugins%2Fopencast%2F%3Fcourse%3D123
```

This Studio Feature is not included in Opencast 8. However, Studio just ignores unknown GET parameters, so it is safe to unconditionally pass those two GET parameters, regardless of Opencast version.

If there are any questions or suggestions, let me know!

ferishili:
With this PR, moodle will pass return params to Studio in order to provide the return button in Studio.
**Admin Settings**
- enable/disable new tab to redirect to studio.
- enable/disable passing return params feature.
- entering dynamic return button label in Studio.
- entering dynamic return url.

**How it work**
- Admin must enable this feature in config setting.
- It passes the url of the page where the request came from (where the Record Video Button is clicked), unless admin configures another URL (relative to wwwroot, capable to have 2 placeholders).
- It passes the site name as default button label, unless admin configures a custom label in the settings.
- Admin are now able to decide whether to use the same tab for Studio or open it on a new tab, by enabling this feature in setting!

### DIFF
--- a/lang/en/block_opencast.php
+++ b/lang/en/block_opencast.php
@@ -666,5 +666,15 @@ $string['editorlticonsumerkey_desc'] = 'LTI Consumer key for the opencast editor
 $string['editorlticonsumersecret'] = 'Opencast Editor Consumer secret';
 $string['editorlticonsumersecret_desc'] = 'LTI Consumer secret for the opencast editor integration.';
 
+// Strings fot new opencast studio settings.
+$string['opencaststudionewtab'] = 'Redirect to Studio in a new tab';
+$string['opencaststudionewtab_desc'] = 'When enabled, studio opens in a new tab.';
+$string['enableopencaststudioreturnbtn'] = 'Show a redirect back button in Studio';
+$string['enableopencaststudioreturnbtn_desc'] = 'When enabled, Studio then renders an additional button "Exit and go back" after up- or downloading the recording.';
+$string['opencaststudioreturnurl'] = 'Custom Studio return endpoint URL';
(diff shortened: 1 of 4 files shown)


TamaraGunkel:
Thanks for the PR!

PR #247, opened 2022-01-13
metadata description with help icon, fixes #183

ferishili:
This PR fixes #183, which shows help icons in front of metadata fields to describe more about them.
**How it works**
- In admin setting page, under _Event Metadata_ and _Series Metadata_ section, the field tables now have another Column called "Field Description".
- If the above field in admin setting has a value, it adds a custom help icon in add forms.
- to render a help icon with custom text, a new method is introduced in `renderer.php` called `render_help_icon_with_custom_text`, which gets a title and content as text and pass them to the `core/help_icon` template to generate a custom help icon.

PR #248, opened 2022-01-18
Initial + scheduled visibility control

### LINKED ISSUE #187: Allow to control the visibility of a video during the initial video upload
As a teacher, I am not able to control the visibility of a video during the initial video upload. I have to wait until the initial video processing is done before I can hide the video if needed. This is especially tricky if I want to upload a video with some assignment solution which should not be public before a defined point in time.

I would appreciate if
a) at least I would be able to set the initial visibility of the video during the initial video upload
b) ideally addittionally a date and time when the video automatically becomes available (which will then require a scheduled task)

ferishili:
This PR fixes #187, which enables teachers to choose the initial visibility as well as scheduling a change visibility for later.

**How it works:**

- Instead of passing all configured roles as acls with initial event upload, the new feature processes the acls based on the selected initial visibility.
- Teachers are able to schedule a visibility change if the required configurations are set. e.g. workflow and allowance!
- The scheduled visibility change then uses a scheduled task and is recommended to be performed every minute, which will read the db records and checks the time and performs the `change_visibility` function from `apibridge`.
- The provided date time for the teacher to set has a default value of 20 minutes after the current time.
- The features are also applied for the ingest upload.


ferishili:
Hi @TamaraGunkel,
the moodle code checker is failing due to errors found in some files that are not included in the commits and are from other places.
Should I fix them as well, or not?
This also applies to the PR #247

TamaraGunkel:
Hey, 
the current master branch does not have any code checker issues.
Could you just rebase your PR so that there aren't issues with unrelated code?

TamaraGunkel:
Okay, weird. Then I'm just reviewing the PR and fixing code issues later if they still arise.

TamaraGunkel:
So, I just had a look at the code and started testing it. So far, everything works great.
However, I'd already like to propose two changes:
- Set the default visibility to 'visible' and not hidden. Teachers are currently used to the fact that their videos are directly available. Changing the default behaviour might be confusing.
- Let the admin configure the minimum waiting time for the scheduled visibility change. 20 minutes are actually quite much, our Opencast instance is usually much faster with processing.

classes/local/addvideo_form.php
```diff
@@ -296,4 +414,34 @@ protected function try_get_string($identifier, $component = '', $a = null) {
...
+     * @param int $ocinstanceid The opencast instance id.
+     * @param array $customminutes Custome minutes to be added or deducted on demand.
+     * @return int
+     */
+    protected function get_scheduled_change_visibility_waiting_time($ocinstanceid, $customminutes = []) {
+        $configwaitingtime = get_config('block_opencast', 'aclcontrolwaitingtime_' . $ocinstanceid);
+        if (empty($configwaitingtime)) {
+            $configwaitingtime = \block_opencast\local\visibility_helper::DEFAULT_WAITING_TIME;
+        }
+        $formtimestamp = $this->formtimestamp;
```
TamaraGunkel, line 430: I'm getting an undefined property error for `formtimestamp` when adding a video. Could you pls fix that?

TamaraGunkel:
I just tried to upload a hidden video. Do I see that correctly that you're not setting any ACLs at all when the video is hidden?

Actually, the "hidden" feature should mean that the video is only hidden for students. The ACLs that are specified as "permanent" in the settings (such as ROLE_ADMIN or ROLE_GROUP_EXTERNAL_APPLICATIONS) should still be set.

ferishili:
Hi @TamaraGunkel, sorry for the bugs. I think, it is ready for your review. Thanks in advance.

ferishili:
2 tests failed, and I cannot find out why, could you please take a look at them?

TamaraGunkel:
So, I tested you changes and it works for me now. Thanks for your work!
The tests just fail sometimes without any visible reasons. Re-running usually fixes that...

ferishili:
No problem, sorry for the small bugs. Thanks a lot ✌️

ferishili:
> Hello, thanks for this feature! I have a question, if as a teacher I configured a wrong date/time by accident, or if I am unsure if I got the date/Time correct, is there any way I can check or change the dates the dates in the surface of the moodle plugin? Regards, Wiebke

Hello,
If I understood your question correctly, you want to edit the start date after upload!
In order to do that you could user update metadata feature, in that form it is now possible to change the start date as well.
BR

wiebkemueller-hsh:
Hello Farbod,

sorry - by saying "feature" I referred to the feature "Initial + scheduled visibility control", I thought this issue was related only to that. Actually I wanted to ask whether one can edit or at least see the date that was set for the change of the visibility status whilst uploading the video after the upload is finished. Is it anywhere in the metadata? Correct me if I am wrong but I think that is not the start date?

BR
Wiebke

ferishili:
Hi Wiebke,
Unfortunately, there is no possiblity to edit the scheduled visibility, but it is possible to change the visibility directly in the overview page after upload and process is done, however, one must pay attention that at the scheduled time the visibility will be changed!
For example, if I upload a video initailly with invisible as its visibility and set it to be visible in 2 days from now. I can then change that visibility as soon as the video is processed. But at the scheduled time (in 2 days) there would be another visiblity change automatically (which I have scheduled)!

PR #254, opened 2022-02-07
User's default values for metadata fields

### LINKED ISSUE #163: Add default values for metadata fields
Certain metadata fields could receive default values:

* title: name of the file
* licence / rights / etc: default per instructor or course
* ...

ferishili:
This PR fixes #163. It is a user driven feature in order to make it easier for every teacher to add its own default values.

**How it works**

- In admin settings page under the metadata fields for both event metadata and series metadata, a new column called "defaultable" is introduced. Admins should check any of the fields as defaultable in order for the users to see the fields in manage defaults page.
- Title in add video page will get the uploaded video name, and it is excluded from the admin settings.
- A new action setting link in index page / A new line of explanation with link in add video / create new series page is introduced, which redirects to the page to manage default values for each user.
- Those explanation links can be redirected back to where they initiated from.
- To store users' default values a new db table called "block_opencast_user_default" is created, which stores userid and the defaults as json string!
- Any default values added inside the manage defaults page, will be inserted as default value to the related fields inside the add video page or create new series dialog.

**All js files are included in the commit as requested!**


ferishili:
Hi @TamaraGunkel,
could you please take a look at the behat tests that are failing, and give me some clue why it throws error? I could not recognize anything problem related to the new codes added!

TamaraGunkel:
Hey, could you integrate the newest master commits and try again?

ferishili:
@TamaraGunkel, I saw that in `tests/behat/block_opencast_manageseries.feature` `metadataseries_1` is lacking. Therefore, instead of defining it inside the behat file which will take a lot of space, I just removed those places that it has been used!

TamaraGunkel:
Thanks! I like the solution with the tables and that it is configurable by the admins.

PR #258, opened 2022-02-10
Remove ACLs when unlinking series

### LINKED ISSUE #225: Unlinking series should change ACL
As with the new import series feature (wich works by adding another series to the course and changing ACL on that series and the videos to grant users on the target course the corresponding permissions), the ACL should be changed back when unlinking a series from a course.

I assume a teacher would expect that students can't access the videos of a series after they unlink the series from the course. Currently this is not the case since the ACL is not changed when unlinking a series.

ferishili:
This PR fixes #225, by changing the visibility of all events to hidden for the course, as well as removing non-permanent course ACLs from series ACLs.

**How it works**

- Introducing a new public function in `apibridge` class `unlink_series_from_course` which is responsible for taking out the course ACLs from series' events as well as series itself.
- The new function will be called upon unlinking series request before deleting the series mapping record.

**Suggestion**

- It would be really nice to have dynamic functions in apibridge class to get from and put back acls into opencast directly, to reduce redundant blocks of codes throughout the class:

1. To get acls dynamically:
```
private function get_oc_acl($where, $identifier) {
        // Get api instance to read acls.
        $api = api::get_instance($this->ocinstanceid);
        $resource = "/api/{$where}/{$identifier}/acl";
        $jsonacl = $api->oc_get($resource);
        $acl = json_decode($jsonacl);

        if (!is_array($acl)) {
            throw new \moodle_exception('invalid_acl', 'block_opencast');
        }

        return $acl;
    }
```
2. To put acls dynamically:
```
private function put_oc_acl($where, $identifier, $acls) {
        // Get api instance to put acls.
        $api = api::get_instance($this->ocinstanceid);
        $resource = "/api/{$where}/{$identifier}/acl";
        $params['acl'] = json_encode(array_values($acls));

        $api->oc_put($resource, $params);
        return ($api->get_http_code() == 200);
    }
```


TamaraGunkel:
Hey,
thanks for your work so far. Everything works as expected.
However, I think that we should delete all course related ACLs. That means, that the ACLs allowing the teachers to access the videos should also be removed. You would need to check the ACLs for course-related identifiers and delete those.
Could you add this?

TamaraGunkel:
As for your remark on the apibridge. Generally, I agree with you that we could refactor some code and make it easier. But I heard, that some people want to develop an Opencast API PHP library and I don't that refactoring the apibridge right now makes sense.

ferishili:
> Hey, thanks for your work so far. Everything works as expected. However, I think that we should delete all course related ACLs. That means, that the ACLs allowing the teachers to access the videos should also be removed. You would need to check the ACLs for course-related identifiers and delete those. Could you add this?
Hey, of course, I am adding it.

ferishili:
Hey @TamaraGunkel,
could you please take a look at failed tests, I could not find any relevance to the changes for this PR!

TamaraGunkel:
Hey,
I've fixed the test, you can integrate them in your PR. 
I'll review the new changes next week.

PR #280, opened 2022-04-27
Multi-Tags and fixing startworkflow!

### LINKED ISSUE #274: workflow_tag works only with one single entry
I tried to set in the moodle Admin-UI for the opencast block two workflow tags, one for an amberscript workfkow for the english  (_amberscript-moodle-eng_) transcription and one amberscript workflow for the german transcription (_amberscript-moodle_), once comma and once semicolon separated. But unfortunately, it works only with one single entry! Neither comma nor semicolon is supported.

Is there any solution?

Thanks,
Michael

ferishili:
This PR fixes #274 as well as the problem with starting the workflow.

**How it works**

- admins are now able to define multiple tags for `Workflow tag` setting using comma-separated list of tags e.g. "upload,archive,amberscript" and so on.
- If multi-tags are defined, plugin will get all workflow-definitions and filter them locally based on them.
- if only one single tag is defined the system will use the previous approach by using the filter query string like "&filter=tag:upload".
- **NOTE**: The start workflow bug has been fixed; the reason was that in mustache template inside message eventlistener parent_origin was not currently evaluated.


mwuttke:
Hello,

I've testet the patch from Farbod - with Moodle 3.10, block_openast v3.11-r7+ and Opencast 11.6 - and the feature works as expected. ;-)

Thanks & Greetings,
Michael

NinaHerrmann:
Hello, thanks for testing :) If you feel confident about this PR, feel free to merge it @Laur0r and I do not have the time to check it.

t-schroeder:
Hi Farbod,

I've tested your implementation and it works. Can I suggest something?

1. The function `get_existing_workflows` still expects a single string as the `$tag` parameter. It then tries to split that string. Just from looking at the function signature (and PHPDoc) it's not clear at all that I can also pass a comma-separated list of tags for `$tag`. Can you make it clearer what values the function supports? Maybe change `@param string $tag` to `@param array $tags` and the do the splitting outside the function, i.e. [here](https://github.com/ferishili/moodle-block_opencast/blob/e4c8456b3aa3911d49073b7e4c1d31ae7723597e/index.php#L50)?
2. For clarity it would be better to have the setting called "workflow_tag**s**" from now on (instead of the original name "workflow_tag"). This require an upgrade step similar to [this](https://github.com/Opencast-Moodle/moodle-block_opencast/blob/eeccbd00baf47cc159b191627c9cdd6788bbce0f/db/upgrade.php#L593-L595) so that existing settings are not lost.

Regards, Tim

ferishili:
Hi Tim, thanks for your review. I will make changes as soon as I got time.

PR #281, opened 2022-04-28
Reading LTI credentials from tool_opencast correctly

### LINKED ISSUE #275: Opencast Studio recordings are not assignd to the series
Opencast Studio can be started and used from the Opencast block as usual. The recordings are also created without errors, transferred to Opencast and processed. But the assignment to the Opencast series of the Moodle course from which the recording was created is not done. Is this error related to the change/move of the LTI settings from the block to the tool plugin?

System used:
Opencast version: 10.11
Moodle version: 3.10
Moodle plugins version: block: v3.11-r8 or -9 & tool: v3.11-r8
OS: Linux Mint
Browser: latest version of Chromium & Firefox

Thanks & Greetings,
Michael

ferishili:
This PR fixes #275 as well as other places that use LTI credentials including (editor and engage redirect).

**Issue**
The problem came from where the calls to read configs from tool_opencast have the ocinstance like` get_config('tool_opencast', 'lticonsumerkey_' . $ocinstance)` where in this case when the oc instance is default it appears to have no `_{instance}`, therefore it gets null in return and refuses to proceed with lti calls!

**How it work**
- There are 2 new functions introduced in apibridge class that get the lti credentials correctly, and then throughout the app we use those functions to get lti credentials.
- There is a new approach to make lti calls in `block_lti_form_handler.js`, which is using AJAX to make the lti first and then redirect to the url that is sent via js_call_amd as parameter.
- The issue with the custom return label for studio is also solved, thanks to the new approach of making lti calls mentioned above.

**NOTE:** The new LTI launch approach with  **AJAX** may require some **server configurations** to overcome the **CROS (Cross-Origin-Allowed)** checkers.

**LATEST UPDATE**
The new approach to perform LTI for Studio is now rejected and won't be applied for this PR anymore, the changes are in the last commit as a correction and will be taken as the final version for this PR.


fishfree:
@ferishili Hi, ferishili! I manually applied your commits  in #275 and this issue, purged all caches, now it shows errors in the console as below:
image
I also have added the add_header instructions in the Nginx reverse proxy long before, and the configuration is right:
image

When I accessed an LTI series added by block_opencast in the course, which is configured here:
image
It shows errors as below:
image

ferishili:
> @ferishili Hi, ferishili! I manually applied your commits in #275 and this issue, purged all caches, now it shows errors in the console as below: image I also have added the add_header instructions in the Nginx reverse proxy long before, and the configuration is right: image
> 
> When I accessed an LTI series added by block_opencast in the course, which is configured here: image It shows errors as below: image
> 
> My block_opencast version is v3.11-r8 (2022032000) , Moodle is 4.0, Opencast is 11.5

Hi @fishfree, please make sure that your consumer key and secret are correct, sometimes copy and pasting the credentials ends up with unwanted whitespaces at the end of the key & secret. The API checker that you send does not check the LTI connection!

ferishili:
> 2\. Why is the new approach for LTI necessary?

Hi @t-schroeder, 
The reason to use this approach is because of providing and consuming the return label and target feature of Studio correctly. The old approach was to use the lti custom_tools that gets an url to be redirected to (in this case studio) after the LTI connection is approved by opencast, that requires the url to be encoded ,however , on the contrast the label accepted by Studio should not be encoded. This phenominal results in error when the return label consists of 2 word separated by (space) e.g. "Moodle Site".

fishfree:
> 

@ferishili I did double check the LTI credentials and there is no problem. Sorry for replying you late.

t-schroeder:
Hi @ferishili,

[this solution](https://github.com/t-schroeder/moodle-block_opencast/commit/c2ab0cd76beecf325db4974806a3d5319aa5e8fd) correctly encodes return labels consisting of 2 words separated by space. I've tested it. So I don't really see why the new approach is necessary. It introduces more code, more steps, potential CORS issues.

ferishili:
> [this solution](https://github.com/t-schroeder/moodle-block_opencast/commit/c2ab0cd76beecf325db4974806a3d5319aa5e8fd) correctly encodes return labels consisting of 2 words separated by space. I've tested it. So I don't really see why the new approach is necessary. It introduces more code, more steps, potential CORS issues.

Hi @t-schroeder,
The solution does not work for me, as it ends up with error in lti (see screenshot)

Screenshot from 2022-06-30 14-49-57

Have you tried to complete all the steps in studio to record the video and check if the button to return with the url and the label appears correctly? The solution will only redirect to the studio without any return button!

t-schroeder:
Hi @ferishili ,

yes, I've tested it. See:

[opencast-studio-multiple-words-return-button.webm](https://user-images.githubusercontent.com/24255259/176860858-421aa471-1e5a-4ebe-828c-da954b03fbfb.webm)

I've used Opencast 10.14 in my tests.

ferishili:
> yes, I've tested it. See:

Hi @t-schroeder 
Though I am still not able to get the result as shown in the video, I am going to revert back the new approach to the last one, and I will create another issue for the Studio itself as this PR meant to solve the the LTI authentication!

taraghi:
Hi

any idea when your patch is released?

PR #286, opened 2022-05-23
Adding get_default_series to web service

BerengarWLehr:
I'm suggesting this addition to the web service for the opencast block. This change is intended to be used in production at the University of Jena to create an automated record-to-moodle workflow that allows teachers to send recorded lectures directly into their (moodle) room series.

### DIFF
--- a/classes/external.php
+++ b/classes/external.php
@@ -105,6 +105,19 @@ class block_opencast_external extends external_api {
         ]);
     }
 
+
+    /**
+     * Returns description of method parameters
+     *
+     * @return external_function_parameters
+     */
+    public static function get_default_series_parameters() {
+        return new external_function_parameters([
+            'course' => new external_value(PARAM_INT, 'course to get series for')
+        ]);
+    }
+
+
     /**
      * Submits the series form.
      *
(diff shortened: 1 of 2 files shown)


ferishili:
Hi @BerengarWLehr,
From where should that external method be called, What I am currently observing is that everything except the call to this external method (namely: "block_opencast_get_default_series") is in place!

That would be great if you could provide a comprehensive description to this PR as soon as possible, in order for me to review and merge it for the upcoming release, otherwise, that will be scheduled for the next one!
THanks in advance.
BR

BerengarWLehr:
With this PR you can record a video in a lecture room and post it into your moodle room using a touch panel where you can search for your moodle course.
The missing link without the PR is to get from the moodle course (which is something a lecturer know in our scenario) to the series the video should be linked into.

ferishili:
How should one test this feature, I would strongly recommend to test it before merge it!

BerengarWLehr:
I think I don't understand your request, what kind of test are you looking for?

ferishili:
I would like to test this feature, as a lecturer? is that possible?

BerengarWLehr:
This addition does not seem to fit to the current code base and is besser suited in another plugin so I close this PR.

PR #289, opened 2022-07-08
Add setting for upload specific timeout

jojoob:
Fixes https://github.com/Opencast-Moodle/moodle-tool_opencast/issues/48

I have implemented this into block_opencast because the upload logic is coded within block_opencast and not tool_opencast.


fsufrre (approved):
works for us

PR #294, opened 2022-09-21
Workaround for replacement problems outside of courses

### LINKED ISSUE #293: Call of /course/management.php fails
When opening `Site-Administration -> Courses -> Manage courses and categories` the call of www.example-server.com/course/management.php fails displaying the following error and stack trace:
    
Screenshot from 2022-09-20 11-42-55

Other than that the plugins are working and the connection test succeeds.

**Block Verison:**      v3.11-r9     2022033100
**Tool Version:**         v3.11-r8    2022032000
**Moodle Version:**    3.9.10+     20210917

berthob98:
Fixes #293

### DIFF
--- a/renderer.php
+++ b/renderer.php
@@ -470,7 +470,8 @@ class block_opencast_renderer extends plugin_renderer_base {
             $html .= $this->output->heading($ocinstance->name);
         }
 
-        if (has_capability('block/opencast:addvideo', $coursecontext)) {
+         global $SITE;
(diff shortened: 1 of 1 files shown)


ferishili:
I am afraid your fix is not enough to prevent this from happening.
In my opinion, we should come up with a fix that:
1. Avoid being added to site administration.
2. Catch and throw errors, that won't break everything!

berthob98:
@ferishili yes I agree I just wanted to provide a quick fix if someone happend to have the same problem

PR #298, opened 2022-10-06
Direct Access to videos' flavors with share button

### LINKED ISSUE #58: Video URLs in overview page

ferishili:
This PR is a new and updated version of the PR #240, which fixes #58,

**Configuration:**

- Admin now has a new setting option right below download channel, called "Direct access channel", by which the opencast publication of the video is served when accessing the video.
- A new access capability is introduced for this feature called "opencast:directaccessvideo".

**Usage:**

- Just like downloading videos, users (lecturers) are now able to see an anchor button in the action menu for each accessible video, which opens a dropdown containing different flavors with different quality of that video.
- Each Direct access link then redirects to a newly introduced page called "directaccess.php". in which the publication url will be extracted and the LTI Authentication will be performed in order to make sure that the access is authorized.

NOTE: TO use in different part, lecturers can now copy the link directly from the dropdown (which peforms the authentication) as well as accessing that link to be redirected to the actual video file on the opencast server and the copy its link and use it based on their needs!


ferishili:
As this PR could not get enough votes to merge, and because keeping the code simpler, this PR will remain open for now and won't be merge into the upcoming release 4.0-r1!

MalteHellwig:
+1

mwuttke:
+1

PR #304, opened 2022-10-18
Transcription

### LINKED ISSUE #107: Subtitles

### LINKED ISSUE #297: download and upload of subtitles
The download and upload of subtitles from the Moodle block of opencast would be a useful possible further feature. For example, to enable post-processing of subtitles that have been automatically created in the course of lecture recordings.

### LINKED ISSUE #303: select the subtitling service and the language when uploading a video
With regard to the issue _download and upload of subtitles_ #297 it would also be a good idea if the teachers could select the subtitling service and the desired language when uploading a video in the opencast block, e.g. Amberscript or Vosk or German or English.

ferishili:
This PR:
fixes #107
fixes #297
fixes #303
replaces #139
It provides a feature to upload attachments alongside the event upload (and in this case transcriptions) via cronjobs as well as managing the transcriptions by providing single uplaod, download and delete.

### Admin Settings

- Under general settings, a new section called "Settings for Transcription" is introduced, which contains required configurations for this feature.
- This feature depends on providing transcription upload workflow, which means if it is empty the whole feature is disabled.
- A custom workflow is needed to delete the transcriptions and if that workflow is not provided the deletion is disabled.
- Service types (Flavors) must be defined as pair of key-value sets as they are being used to provide the dropdown selection in the ui later on.
- Download transcriptions can be perform with or without LTI auth, which is meant to be used when Secure Static Files is enabled.

### Usage (Teacher)
- Teachers will now have an additional section in the add video page called "Upload Transcription Files", in which they will the trasncription sets (number of sets can be defined in admin settings).
- A transcription set contains of 2 parts: 1. Service Type (Flavor) 2. the file (mostly with .vtt extension)
- When uploading via add video page, then the whole attachments set will be recorded into a db table, and a cronjob will supports that procedure to check (by default every minute) to first check if the uploading and processing the video is completed, if so, it proceeds with adding the attachment files via ingest and starts the configured workflow.
- On the other hand, in overview page, the video table will now have another items in the action menu called "Manage Transcriptions", by which teachers will be redirected to a new page to view the list of current transcriptions, download one (supports LTI), delete one and add new one.
- Single Transcription Upload will use direct upload (add new attachment)
- Single Download will download the transcription file directly or if the admin setting is set to perform LTI before download, it redirects the teacher to the actual file due to performing LTI, from there they can copy and paste the content.
- Single Delete: Note that this action requires a custom workflow, by which the workflow must look into the newly update caption list and process them, at the end publish them into engage and so on.
- NOTE: Uploading a new transcription with the service type that has been used before, results in the replacement of that transcriuption with the lates uplaod, which is meant to update a single transcription. 

### To test this PR:
you will need:
- The latest master version of [tool_opencast](https://github.com/Opencast-Moodle/moodle-tool_opencast/tree/master) since it has the feature of uploading files other than videos.
- An upload workflow for captions must be configured in opencast. please refer to this [comment](https://github.com/Opencast-Moodle/moodle-block_opencast/issues/107#issuecomment-592621749)
- HTML track files with .vtt
- To delete a custom workflow must be used.
- Defining the related admin setting first.
- Moodle cronjobs: \block_opencast\task\process_attachment_cron must be active and running.
- Purging all caches in moodle is recommended.


ferishili:
A simple default workflow that can be used for both upload as delete of transcription is as follows:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<definition xmlns="http://workflow.opencastproject.org">

  <id>republish-metadata-and-captions</id>
  <title>Republish metadata and captions</title>
  <tags>
    <tag>api</tag>
  </tags>

  <operations>

    <!-- Tagging Captions -->

    <operation
      id="tag"
      description="Tagging Captions">
      <configurations>
        <configuration key="source-flavors">captions/*</configuration>
        <configuration key="target-tags">+archive</configuration>
      </configurations>
    </operation>

    <!-- Taking a snapshot -->

    <operation
      id="snapshot"
      description="Archive current captions">
      <configurations>
        <configuration key="source-tags">archive</configuration>
      </configurations>
    </operation>

    <!-- Publish to engage player -->

    <operation
      id="publish-engage"
      exception-handler-workflow="partial-error"
      description="Update recording in Opencast Media Module">
      <configurations>
        <configuration key="download-source-flavors">dublincore/*,security/*,captions/*</configuration>
        <configuration key="strategy">merge</configuration>
	<configuration key="merge-force-flavors">dublincore/*,security/*,captions/*</configuration>
        <configuration key="check-availability">false</configuration>
      </configurations>
    </operation>

    <!-- Clean up work artifacts -->

    <operation
        id="cleanup"
        fail-on-error="false"
        description="Remove temporary processing artifacts">
      <configurations>
        <configuration key="delete-external">true</configuration>
        <!-- FixMe Don't clean up ACLs until workflow service no longer looks for them in the WFR. -->
        <configuration key="preserve-flavors">security/*</configuration>
      </configurations>
    </operation>

  </operations>
</definition>
```

mwuttke:
I've tested the new feature witch the mentioned workflow and it works in our test setup as expected with Moodle 3.11.10+, Opencast 11.11 and the latest opencast tool plugin (master branch).

@t-schroeder & @ferishili: Thanks a lot for your work!

Greeting,
Michael

PR #305, opened 2022-10-21
Live Status Update

### LINKED ISSUE #180: Show position in OC processing queue / show remaining estimated processing time
If a teacher uploads a videos to Opencast through block_opencast, they are completely uninformed how long the processing of the video will take. They just see the orange clock icon and would basically have to reload the page and check the status again and again until to be informed when the video is finally available.

While #111 might reduce this intransparency, I would like to propose to add a processing queue information to the video overview table within /blocks/opencast/index.php.

Possible useful information for teachers could be:
1. Directly after the upload to Opencast:
"Processing of this video is queued, but has not yet started. The video is currently waiting behind 20 other videos in the processing queue."
2. As soon as processing has started:
"Processing of this video has started. The process has 40 processing steps and currently 20 of these steps are finished."
or 
"Processing of this video has started. Processing a video of this length approximately takes 40 minutes which means it should be done more or less in 20 minutes".

I am aware that the way how Opencast works does not allow to give reliable figures for processing times. However, as described in the examples above which should be realizable and will help teachers more than not knowing anything currently.

ferishili:
This fixes #180 by providing info about upload queue as well as introducing a new live update mechanism to update the status of every ongoing active processing-event in a one second interval.

### Upload Notification
A new uploading notification is added to the existing event status notification system, which notifies the course teachers as well as the uploader about the upload queue (right after the add video is done and the upload job is waiting in queue to get uploaded and transffered.)

### Live status update
Initially the request demanded a notification that could inform teachers how long should the processing of an event take until it is completed, however, that could only be possible when the workflow instance id is provided in the system, which in this case it is only possible via ingest upload. (Also the GET:/api/workflows endpoint is deleted for newer version of OCs). With that said, the workflow processing duration time estimation is **not possible at the moment**.
_Instead of providing such information a new mechanism to update the status live is introduced:_

- The live status update will be triggererd when there is one or more flags (those events or upload jobs that are currently being processed).
- It performs an Ajax call to its service in a 1 second interval by which the current state of the processing is evaluated.
- When the workflow/upload process gets to the point that is either failed or succeeded, it just updates the frontend and then remove the flag.
- When an event is considered as not processing state, then a reload will occure in order for the backend to to its job for providing the new status features to the frontend.
- The interval would be cleared when there is no flag identified.

###Usage
With the help of this feature teacher now are being notified about uploading queue and then they do not need to refresh the index page to check the processing status anymore.

###To test this PR
You would need to have:
- latest tool_opencast master branch installed.
- Admin configuration for this feature called "Live Status Update" under additional features must be configured.
- All moodle caches must be purged.


mwuttke:
I've tested the new feature and it works in our test setup (with Moodle 3.11.10+, Opencast 11.11 and the latest version of the opencast tool plugin on master branch) as expected.

@ferishili: Thanks a lot for your work!

Greeting,
Michael

PR #314, opened 2023-01-22
Autocomplete Suggestions using capability

### LINKED ISSUE #302: Hardcoded role name
For the "Presenter(s)" and "Contributor(s)" autocomplete fields we get the users in the current course that have the role with shortname "editingteacher": https://github.com/Opencast-Moodle/moodle-block_opencast/blob/653e01ea8e774c695c65b2234fa55ba926f81e0c/classes/local/autocomplete_suggestion_helper.php#L64-L71
In our Moodle instance there is no role with that name. So this always creates a warning in the following lines when trying to access `$role->id` since `$role` is `false`: https://github.com/Opencast-Moodle/moodle-block_opencast/blob/653e01ea8e774c695c65b2234fa55ba926f81e0c/classes/local/autocomplete_suggestion_helper.php#L71-L75

In my opinion there should be no hardcoded role. It should either be an admin setting where you select a role (or multiple roles) or it should be done through a capability.

ferishili:
This PR fixes #302,

### Description
The autocomplete suggestions for contributor and presentor fields consists of 3 different part, one of them is to extract the teachers name enrolled in the course, this process had a few problems along the way, one was that it was using shortname "editingteacher" to get the role id which was incorrect as it is a parameter that could be changed, and for those institutes that have changed this parameter, this process throws error and does not work properly.

### Solution
A new capability is introduced to provide appropriate roles defined by "archetypes". And on the other hands the process will now use a built-in method "get_enrolled_users" from Moodle to get the enrolled users who have that capability.

### To Test and run this PR
- An upgrade to a new version of this plugin is needed (Automatically). Because a new capability is introduced in `access.php` file, therefore, to test/run this PR one should pay attention to this version bump, to avoid versioning confilct etc.
- Make sure the test course have a few teachers enrolled.
- In update metadata / add video / series add or update pages you have be able to see the teachers' fullnames as autocomplete suggestions dropdowns for Presenter and Contributor fields.


ferishili:
Merged in [b2f7559aaef5752889872a8e2b4569b04fda675d](https://github.com/Opencast-Moodle/moodle-block_opencast/commit/b2f7559aaef5752889872a8e2b4569b04fda675d) because the github was down, it did not follow the normal way of mergin PRs.

PR #326, opened 2023-04-04
fixed record video button

aliyevTU:
fixed record video button redirecting to /lti instead of /studio


ferishili:
Hi @aliyevTU,
May I know the reason, why you came up with this PR?
To make it more clear, the mechanism of LTI to be performed and be processed by Opencast is to no matter where you want to land in opencast, you have to make the call to /lti first.
Then Opencast processes this LTI call with the all the parameters you passed, if it is successfull, then it redirects to the desired place passed as customtool!

ferishili:
I close this PR as it is invalid

PR #338, opened 2023-07-06
New query param to be appended into Studio url path

### LINKED ISSUE #337: add configuration field for settingsFile Parameter
To get the patch from lukas https://github.com/elan-ev/opencast-studio/pull/1040 work well with moodle it would be nice to have a certain field at moodle block PlugIn for the SettingsFile Parameter.
As an entry you can set the correct path to the settingsFile Query Parameter: For instance: "(https://my.edu.org/studio?settingsFile=settings-moodle-A.toml)
Same way as for the configuration field for return.label

ferishili:
This PR fixes #337 

### How it works
- a new method is introduced in `apibridge` class called `generate_studio_url_path` to have a central place responsible for that.
- a new setting called `opencast_studio_custom_settings_filename` is provided in admin setting to enter the custom setting FILENAME for the studio.
- In case of redirect to studio without lti call, the label and return url + setting filename is now also appended to the studio url path.

### To test this PR:
You would need:
- latest version of tool_opencast master branch.
- the changes in Opencast side will be in place and functional only in Opencast v14.2
- After installing this PR make sure you purge the caches for everything especially texts to work properly.

**NOTE**: As of writing this PR, it is not yet tested and finalized, therefore upcoming updates and improvements will be expected. Keep track of the commits adding to this PR!

### DIFF
--- a/classes/local/apibridge.php
+++ b/classes/local/apibridge.php
@@ -2957,4 +2957,84 @@ class apibridge {
         }
         return false;
     }
+
+    /**
+     * Central place to generate studio url path and its query params if applicable.
+     *
+     * @param string $courseid Course ID
+     * @param string $seriesid Series identifier
+     * @return ?string the generated studio url path or null if something went wrong.
+     */
+    public function generate_studio_url_path($courseid, $seriesid) {
+        global $SITE;
+        // Return null if the requirements are missing.
+        if (empty($courseid) || empty($seriesid)) {
+            return null;
+        }
+        $studiourlpath = '/studio';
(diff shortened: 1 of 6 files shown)


fsufrre:
I tested this PR with Opencast 14.0 and the Patch from https://github.com/elan-ev/opencast-studio/pull/1040.
I sucessfully made it to run one Moodle 3.11 instance and two 4.1 instances to work fine with Opencast Studio and the correct LTI-Roles. I would happily suggest to merge this to the upcoming release.

PR #339, opened 2023-07-19
Fix cutting off action menue items in overview video table 

### LINKED ISSUE #261: Download options box cut off
image

This happens when the Download option box exceeds the area of the series.

Bildschirmfoto vom 2022-02-25 12-21-04

ferishili:
This PR fixes #261

### How it works:
- in recent versions of Moodle there would be a div with `no-overflow` class before any flexible table rendered, by this PR with override the behavior of that div!

### To test:
- Latest version of tool_opencast
- Make sure you have purged all the caches after installing this patch


ferishili:
This problem is in fact a known bug in Moodle, following this: [MDL-77375](https://tracker.moodle.org/browse/MDL-77375)

That means, the initial fix commits of this PR are no longer needed. And since the affected Moodle versions are (4.0.6, 4.1.1, 4.1.2, 4.1.3, 4.2 regressions, 4.2, 4.2.1, 4.2.2), one must upgrade his/her Moodle to [4.1.5](https://tracker.moodle.org/issues/?jql=project+%3D+MDL+AND+fixVersion+%3D+4.1.5), [4.2.2](https://tracker.moodle.org/issues/?jql=project+%3D+MDL+AND+fixVersion+%3D+4.2.2) which at the time of writing this are not yet released!

What is left to do from our side is to only use `set_boundary` function when creating action menus! And hope everyone sees this PR and gets informed about this **MOODLE BUG** (not `block_opencast`!) That they need to upgrade their moodle to above-mentioned-fixed versions and not creating issues reporting a bug related to this repository!

ferishili:
For Info: the fixed moodle releases (4.1.5 & 4.2.2) are now (as of writing this) available!

PR #347, opened 2023-11-15
Update for Moodle 4.3

justusdieckmann:


ferishili:
An important concern here!
This PR should **not** been merged **without** 1. green test checks 2. a review from others!

PR #348, opened 2023-11-16
More accurate LTI module cleanup process

### LINKED ISSUE #346: Deletion of other external lti tools
We experienced a problem where **block_opencast** deletes the database entries of other external lti tools.

release = 'v4.1-r1';
version = 2023052600;

The block wrongfully deletes entries in the **mdl_lti** table in the moodle database.
We include [Matlab Grader](https://de.mathworks.com/help/matlabgrader/ug/integrate-matlab-grader-with-moodle-lti-1-3.html) as an external lti tool, thereby entries in the **mdl_lti** table are written.

The regular execution of the [block_opencast\task\cleanup_lti_module_cron](https://github.com/Opencast-Moodle/moodle-block_opencast/blob/d608ec275126874aa471f8dd544d66ea06704425/classes/task/cleanup_lti_module_cron.php) task leads to the deletion of these entries.

More specific the [cleanup_lti_module_entries](https://github.com/Opencast-Moodle/moodle-block_opencast/blob/master/classes/local/ltimodulemanager.php#L1304) function deletes these entries in line 1369 and 1371.

I didn't dig deep enough in to the purpose of the task yet, but to my understanding there should be a check preventing the deletion of non opencast lti modules.

I hope this is helpful and my understanding of the problem is correct thereby preventing others from experiencing this problem.

ferishili:
This PR fixes #346

#### How it works
- Basically it checks if the related setting for specific `ocinstance` is enabled and proceeds, otherwise, removes the opencast lti module record from the table to avoid unwanted behavior.
   - `ltimodulemanager::update_existing_lti_modules` -> checks and decides whether to remove or proceed with its job.
   -  `ltimodulemanager::record_manually_added_lti_modules` -> checks and decides whether to remove or proceed with its job
   -  `ltimodulemanager::cleanup_lti_module_entries` -> refuses to delete the records if the setting is disabled.

**NOTE:** Please try to review the code and conduct an in-depth test, before merge!


NinaHerrmann:
Could you take a look at the failing test and write an additional test regarding the feature before requesting the review? :)

ferishili:
> Could you take a look at the failing test and write an additional test regarding the feature before requesting the review? :)

The failing test is only a timing mismatch and will succeed with a re-run!
Additional test step is not needed!

NinaHerrmann:
Why is no test needed? As it caused a bug previously I would expect a test checking if after deleting an opencast instance the number of extern tools is equivalent.

ferishili:
I am not 100% sure, if it was a bug, and it is only a double checker!

ferishili:
The checks are all green!
Let me see what I can do about additional feature steps, however, it would take some time!
Thanks!

berthob98:
Thank you for the fix @ferishili
I experienced one error when testing

### Problem
With this patch applied running the task [block_opencast\task\cleanup_lti_module_cron](https://github.com/Opencast-Moodle/moodle-block_opencast/blob/d608ec275126874aa471f8dd544d66ea06704425/classes/task/cleanup_lti_module_cron.php) produces an error for me.
**Log:**
```
Step 2: Looking for manually added LTI modules...
... cleanup process failed:
Fehler in der Kodierung gefunden, den nur Programmierer/innen korrigieren können: moodle_database::get_in_or_equal() does not accept empty arrays
```
### Analysis
I think this line 
`        list($insqltoolids, $inparamstoolids) = $DB->get_in_or_equal(array_column($toolids, 'id'));`
in [classes/local/ltimodulemanager.php](https://github.com/Opencast-Moodle/moodle-block_opencast/blob/93448e39a7552646b6c9e56bcad4626c6f30db1d/classes/local/ltimodulemanager.php) doesn't like the empty **toolids[]** array, from skipping every entrie in the [lines before.](https://github.com/Opencast-Moodle/moodle-block_opencast/blob/93448e39a7552646b6c9e56bcad4626c6f30db1d/classes/local/ltimodulemanager.php#L1183)
### Cheap Fix
My first Idea was to skip the totality of these 3 functions
- record_manually_added_lti_modules()
- update_existing_lti_modules()
- cleanup_lti_module_entries()

if both **addltiepisodeenabled_** and **addltienabled_** are disabled. Not having to deal with problems of that sort. If one of these is two is enabled the error shouldn't occure, because at least one **preconfigured_tool** will exists. But maybe there are better ways to handle this.

ferishili:
> if both **addltiepisodeenabled_** and **addltienabled_** are disabled. Not having to deal with problems of that sort. If one of these two is enabled the error shouldn't occure, because at least one **preconfigured_tool** will exists. But maybe there are better ways to handle this.

Thanks for your testing and analysis. You are right, escaping the functions when they are not enabled, is one way to do, however, it compromises the whole concept and idea behind this feature and also the multi-instance is not supported!

ferishili:
> Why is no test needed? As it caused a bug previously I would expect a test checking if after deleting an opencast instance the number of extern tools is equivalent.

To make things clear, it is not about deleting an opencast instance and check if external tool exists, but only to escape the lti cleanup feature when the setting is deactivated.

ferishili:
@justusdieckmann was there a reason you have disabled the behat feature? I have noticed that, moodle 403 does not like the old way.

ferishili:
@berthob98 would you please test the latest commit again.

justusdieckmann:
> @justusdieckmann was there a reason you have disabled the behat feature? I have noticed that, moodle 403 does not like the old way.

@ferishili Yes, it seems that manually specifying custom parameters for lti tool activities is not possible anymore in 4.3 because of [MDL-78916](https://tracker.moodle.org/browse/MDL-78916). That means we can't manually create opencast lti activities, so I had to deactivate the behat feature that tests creating those. I still want to take a closer look, and maybe start a discussion, but since we want to get this release out as soon as possible, I've thought it best to postpone that until after the release :)

ferishili:
> @ferishili Yes, it seems that manually specifying custom parameters for lti tool activities is not possible anymore in 4.3 because of [MDL-78916](https://tracker.moodle.org/browse/MDL-78916). That means we can't manually create opencast lti activities, so I had to deactivate the behat feature that tests creating those. I still want to take a closer look, and maybe start a discussion, but since we want to get this release out as soon as possible, I've thought it best to postpone that until after the release :)

All right, thanks for the reply.
Following the request from @NinaHerrmann to provide behat test features, and according to the test results that were green (except in 4.3), we are now safe to merge this PR.
PS: the behat test feature now contains the scenario to cover the problem!

NinaHerrmann:
As we are planning the 4.3 Release it is not an option to let the 4.3 test fail...

NinaHerrmann:
I also feel like 70 % of Pull Requests are just replaceable discussions regarding testing etc. :joy: 
Sorry, I understood you wrong! Looking forward to the ci-results, thank you for all the work - although I write sometimes a little bit unfriendly, I really appreciate it!!!

ferishili:
@NinaHerrmann All GREEN, you are good to go!

NinaHerrmann:
Thank you so much! :+1:

PR #349, opened 2023-11-22
Using moodle bautifier to fix at least the most obvious problem 

NinaHerrmann:
- hope it doesn't break anything...


ferishili:
Hi @NinaHerrmann,
the moodle phpdocs is not happy with the followings, take care of them and you will be fine:
Screenshot 2023-11-22 at 15 31 40
Screenshot 2023-11-22 at 15 31 48

PR #354, opened 2023-12-21
Proposal: (delete block) removing seriesmapping with new delete action link 

### LINKED ISSUE #351: After deleting and re-adding an Opencast block, there are two Opwencast series with the same name

ferishili:
This PR is a form of proposal that would fix #351

### How it works
- A new action item is introduced that indicated the delete of series mapping as well.
- The block delete action items now has more descriptive title and information.

NOTE: the wording may need to be adjusted and prettified.

### Tutorial Video

https://github.com/Opencast-Moodle/moodle-block_opencast/assets/53179227/524468ab-48a6-42bf-9e8e-eaafbde01e84


ferishili:
Close as completed in https://github.com/Opencast-Moodle/moodle-block_opencast/pull/355

PR #355, opened 2023-12-21
Proposal: (delete block) removing seriesmapping with confirmation 

### LINKED ISSUE #351: After deleting and re-adding an Opencast block, there are two Opwencast series with the same name

ferishili:
This PR is a form of proposal that would fix #351

### How it works
- The block delete action item now has more descriptive title and information.
- It consists of 2 step:
   - 1: The default modal that user is able to make sure that she/he wants to delete the block.
   - 2: The confirmation to deleting the series mapping as well: in this confirmation page, user is able to decide whether to delete the series mapping as well.

NOTE: the wording may need to be adjusted and prettified.

### Tutorial Video

https://github.com/Opencast-Moodle/moodle-block_opencast/assets/53179227/08a9b7d5-30a4-4eb4-8db6-10cf4e1cec9e


justusdieckmann:
Hey @ferishili, my apologies for the delay.

Thank you very very much for your hard work and giving us options to choose from! We like this approach, but would rather have one popup instead of two.

I created a very quick (non-functional) popup prototype to see if the idea was any good:
image

What is your oppinion? I would gladly add a commit to this PR to merge the two popups. If you would like to do the remaining programming instread, I can push my (very much) WIP commit somewhere.

Thank you again!
Justus

ferishili:
> Hey @ferishili, my apologies for the delay.
> 
> Thank you very very much for your hard work and giving us options to choose from! We like this approach, but would rather have one popup instead of two.
> 
> I created a very quick (non-functional) popup prototype to see if the idea was any good: image
> 
> What is your oppinion? I would gladly add a commit to this PR to merge the two popups. If you would like to do the remaining programming instread, I can push my (very much) WIP commit somewhere.
> 
> Thank you again! Justus

Hi @justusdieckmann, hi @NinaHerrmann,
Happy new year!
No problem at all and thanks for your feedback.
About adding options as button in that dialog, I am not sure if it could follow the moodle standards, or we need to provide some hacky-workarounds! And I assume that we have to follow Moodle Standards according to @NinaHerrmann!
However, please go ahead and use this PR and add/change/enhance and make your changes.
Thanks

justusdieckmann:
Hey,

happy new year to you too! This would be my attempt.

[Screencast from 11.01.2024 19:03:49.webm](https://github.com/Opencast-Moodle/moodle-block_opencast/assets/45795270/8eab0418-3218-472d-89b0-c7720be40eda)

It requires the `manageseriesforcourse` capability to be shown the popup.
In #356 we change the capability so that teacher have that by default on new installations. Also we will suggest giving teachers that capability in the next release notes. 

I used the unlink_series external-api call to remove the series from the course before deleting the block, since it seemed to be the way with the least amount of extra boilerplate code, but I could of course also do it in a different way.

What is your opinion on the general approach? (The ci on the pr did pass, but the one on the commit didn't, so those seem to be random connectivity failures, but I somehow don't have permission to restart the workflows. Will do something about that tomorrow :)  )

Cheers
Justus

ferishili:
> Hey,
> 
> happy new year to you too! This would be my attempt.
> 
>  Screencast.from.11.01.2024.19.03.49.webm 
> It requires the `manageseriesforcourse` capability to be shown the popup. In #356 we change the capability so that teacher have that by default on new installations. Also we will suggest giving teachers that capability in the next release notes.
> 
> I used the unlink_series external-api call to remove the series from the course before deleting the block, since it seemed to be the way with the least amount of extra boilerplate code, but I could of course also do it in a different way.
> 
> What is your opinion on the general approach?
> 
> Cheers Justus
Hey,
Thanks for your work, well done, that is a good improvement.
This seems good to me and could fulfil the need in one modal.

BR

PR #356, opened 2024-01-08
Make seriesmanagement work properly with or without capabilities to create or import series.

justusdieckmann:
Removes the defineseriesforcourse capability, since it wasn't coherently used.

Now requires manageseriesforcourse for `manageseries.php`, and show import series / create new series buttons depending on respective capabilities.

Give teachers manageseriesforcourse and createnewseries capabilities by default on new installations.


ferishili (approved):
Thanks for the work, that looks good to me and since `behat` tests are all passed it is good to be merged.
As you said, the documentation about this change must also be updated, and admins must be aware of this change since it is privacy-related.

PR #358, opened 2024-02-21
Batch/Bulk Upload

### LINKED ISSUE #350: Suggestion: Add multi- / bulk upload feature
We have lecturers which upload quite often a bulk of videos. (Sometimes even 20 up to 40 pcs at once.)

Single uploads are therefor quite a pain / take away alot of time. (As you have to wait for the single video to upload.
And only after it finished you can continue again to edit the metadata of the next one + upload again.)

<ins>Great would be therefor:</ins>

- Bulk upload for at least a few videos at once (maybe a limited amount if needed for whatever reason)
- Upload with the names of the files as titles (or editing of at least the titles of the events per video entry)
- MAYBE keep the system as it is but add a queue feature which you can trigger then at once so its getting processed one after each other

Would be great to have a bulk upload option of any kind.
Btw.: Thx for your hard work and greetings from Austria.

ferishili:
This issue fixes #350,

### Description
please refer to the issue,

### How it works
- A new setting is provide to activate/deactivate the feature "Enable batch video upload".
- A new button called "Add videos (batch)" is provided in overview page as well as block sidebar component, which redirects to the batch upload page.
- In order to provide any event metadata, you should check the `Batchable` option of the desired metadata in the Event metadata settings.
- Only Presenter videos can be uploaded into the file manage component in the batch upload page.
- The title of the file will be selected as for the title of the video.
- One could drag and drop files in the filemanager component or select multiple files one at a time.
- The uploaded files will get a upload job record (sit in the upload queue) to be picked up and transferred to the Opencast.

### Important
- You need consider increasing the upload job limit as well as timeouts for upload and the Opencast API instance setting.


taraghi:
Hi @ferishili 


We had the honor of testing this patch at Graz university of technology, running on Moodle 4.1.9 with PHP 7.4 

I, as manager, switched my role to another user role in a course with block opencast added containing a series.
Then I switched back my role to the manager, and got the following error message

Bildschirmfoto 2024-02-26 um 15 27 47


The php function `str_contains()` exist since PHP 8. As we have PHP 7.4 running, this leads to a PHP error. There are alternative php functions such as `strpos()` that can be used instead, in order to support lower PHP versions.

Thank you!

ferishili:
Hi @taraghi,
thanks for your feedback, it is indeed a backward incompatibility of PHP versions and has to be fixed. However, it requires a new (separate) issue, because it is not related to this PR and its content!
Would you mind creating that issue, I will then proceed with the fix ASAP.
Thanks in advance.

llttugraz:
Hello @ferishili,
when testing this the following behaviour has been observed:
_When using the bulk upload, you have to load all the videos into the one upload window. You can drag and drop several videos and it works. However, if you do not use drag and drop but the file browser, you can only select one file at a time and have to choose all of them individually, which is confusing (expected behaviour would be that you can choose several files inthe file browser with shift or command and upload them in bulk as well)._

Should we open a new issue for this?

Thanks,
y

taraghi:
> Hi @taraghi, thanks for your feedback, it is indeed a backward incompatibility of PHP versions and has to be fixed. However, it requires a new (separate) issue, because it is not related to this PR and its content! Would you mind creating that issue, I will then proceed with the fix ASAP. Thanks in advance.

Hi @ferishili 

As far as I see the specified php function is used in this patch. It is not used in the last stable version. But I can for sure write a separate issue for that. No problem.

ferishili:
> > Hi @taraghi, thanks for your feedback, it is indeed a backward incompatibility of PHP versions and has to be fixed. However, it requires a new (separate) issue, because it is not related to this PR and its content! Would you mind creating that issue, I will then proceed with the fix ASAP. Thanks in advance.
> 
> Hi @ferishili
> 
> As far as I see the specified php function is used in this patch. It is not used in the last stable version. But I can for sure write a separate issue for that. No problem.

Hi,
the php 8 function was introduced here: https://github.com/Opencast-Moodle/moodle-block_opencast/blob/8c00eec04409e0c09bd10182fbf1f2cba894555b/block_opencast.php#L217
via this commit: https://github.com/Opencast-Moodle/moodle-block_opencast/commit/8c00eec04409e0c09bd10182fbf1f2cba894555b

And of course, it is in master not in the latest stable version!

Since we are at it, the following php 8 functions will also make problem:
`str_starts_with`: https://github.com/Opencast-Moodle/moodle-block_opencast/blob/8c00eec04409e0c09bd10182fbf1f2cba894555b/block_opencast.php#L228

ferishili:
And the reason, I insist on having a separate issue, is just to have a clear issue to PR relationship and not fix/add something else other than the context described in the issue #358. In this way, future developers do not have a hard time debugging or looking for something when referring to the git history.

taraghi:
@ferishili  I see.  Thank you very much for clarification.

I just wrote a separate issue including also your comments: https://github.com/Opencast-Moodle/moodle-block_opencast/issues/359

ferishili:
Hi @taraghi,
the incompatibility problem is now fixed and is also merged from master to this PR. Please give it another try after the CI checks are all green.
Thanks & BR

llttugraz:
@taraghi, it all looks green :)

ferishili:
Hi @taraghi, @llttugraz,
This PR is still waiting for your review, please make sure that it does wait much longer.
Thanks in advance,
BR

batchupload.php
```diff
@@ -0,0 +1,306 @@
...
+    $error = null;
+    $totalfiles = count($batchuploadedfiles);
+    // Loop through the files and proceed with the upload and cleanup records.
+    if (!empty($batchuploadedfiles)) {
+        $errorcount = 0;
+        foreach ($batchuploadedfiles as $uploadedfile) {
+            try {
+                if ($uploadedfile->get_filename() === '.') {
+                    file_deletionmanager::fulldelete_file($uploadedfile);
+                    $totalfiles--;
```
justusdieckmann, line 248: I might have overlooked something, but this isn't needed, right?

classes/local/batchupload_form.php
```diff
@@ -0,0 +1,413 @@
...
+
+use block_opencast\opencast_connection_exception;
+use block_opencast_renderer;
+use coding_exception;
+use core\notification;
+use html_writer;
+use local_chunkupload\chunkupload_form_element;
+use moodle_url;
+use moodleform;
+use MoodleQuickForm;
```
justusdieckmann, line 36: chunkupload_form_element and MoodleQuickForm are unused

classes/local/batchupload_form.php
```diff
@@ -0,0 +1,413 @@
...
+ * @author     Farbod Zamani Boroujeni <zamani@elan-ev.de>
+ * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+class batchupload_form extends moodleform {
+
+    /**
+     * Form definition.
+     */
+    public function definition() {
+        global $CFG, $DB, $PAGE , $OUTPUT;
```
justusdieckmann, line 58: $CFG and $OUTPUT are unused

justusdieckmann (changes requested):
Hey @ferishili,

The pull request looks good to me in general. I have added review comments for a few bugs and typos I found and some nit-picking. Please see what you think of them :)

Cheers
Justus

version.php
```diff
@@ -24,8 +24,8 @@
 defined('MOODLE_INTERNAL') || die();
 
 $plugin->component = 'block_opencast';
-$plugin->release = 'v4.3-r1';
-$plugin->version = 2023112200;
+$plugin->release = 'v4.3-r1-i-350';
```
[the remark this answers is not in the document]
NinaHerrmann: I would prefer not to change the release name at all. In case we make a release we do it in the commit for the Release.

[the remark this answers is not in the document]
ferishili: Yes, of course, that I missed, and it was only for my test purposes!

llttugraz:
@ferishili, thanks very much :)
It all looks good on our end. We can provide the German translation strings if you could include them — unless @taraghi we can add them directly in the code(?)

ferishili:
> @ferishili, thanks very much :) It all looks good on our end. We can provide the German translation strings if you could include them — unless @taraghi we can add them directly in the code(?)

Happy to hear that, thanks for the feedback. I will proceed with change requests and merging it.
PS: The plugin has to have only English translation (default) in the master, other translations need to be added via external locale tools that Moodle offers, which I don't know much about (You need to ask the community, or event better @mwuttke who has the most experience 🥇 )

batchupload.php
ferishili: You mean the whole '.' file name checker?
If yes, it is necessary here!

batchupload.php
```diff
@@ -0,0 +1,306 @@
...
+        if (!$apibridge->is_owner($ocseries->acl, $USER->id, $SITE->id)) {
+            redirect(new moodle_url('/blocks/opencast/overview_videos.php', ['ocinstanceid' => $ocinstanceid,
+                'series' => $series, ]),
+                get_string('addvideonotallowed', 'block_opencast'), null,
+                notification::NOTIFY_ERROR);
+        }
+    }
+} else {
+    $coursecontext = context_course::instance($courseid);
+    require_capability('block/opencast:addvideo', $coursecontext);
```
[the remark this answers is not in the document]
ferishili: In case that series has specific value is only when the form is submitted via POST, therefore you will never be able to post a form with any arbitrary series (unless you are a hacker :D)

Therefore, we leave this as it is!

ferishili:
Hi @justusdieckmann,
thanks for the review, the change requests have been resolved, please have a look at them, I answered some of them directly if not changed!
BR

batchupload.php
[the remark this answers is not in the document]
NinaHerrmann: We also want to be protected against any type of unwanted behaviour, which is why it is necessary to include the check.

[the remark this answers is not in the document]
ferishili: Done in the last commit, please wait until all test are completed

batchupload.php
justusdieckmann: No, sorry, I just meant the `$totalfiles` variable and decrement

ferishili: The `$totalfiles` variable is used for success info and decrement is also important to show the accurate file count result!

justusdieckmann: You're right, I somehow overread that. Thanks!

batchupload.php
```diff
@@ -0,0 +1,322 @@
...
+
+    // Make sure apibridge is available to use.
+    if (!isset($apibridge)) {
+        $apibridge = apibridge::get_instance($ocinstanceid);
+    }
+
+    // Double check if the series is related to the course, which never happens, but we check anyway.
```
justusdieckmann, line 123: This security fix was very important to me as well. It's not certain that only "real" teachers have the permission, for example students who were tutors once might have it as well and every computer science student is probably to some degree a 'hacker'/a bit curious what they can do in a system (One doesn't have to be especially malicious to look through this code and test out if that means you can really upload videos to any series. Maybe they would even report that issue afterwards). And even if students should never have this capability, I really wouldn't be comfortable leaving it as it was, because someone could gain that capability by fault, and in general, we don't want to rely on users (even more privileged ones) to behave nicely.

Things like these "never happen", until they do happen. :) And we want a robust plugin where we are certain that it can't possibly happen.

You could of course only accept the `series`-param if `$courseid == $SITE->id`, if you'd like that better than checking whether the series is connected to the course.

settings.php
```diff
@@ -262,6 +264,22 @@
                 get_string('maxseriesdesc', 'block_opencast'), 3, PARAM_INT
             ));
 
+            // Batch upload setting.
+            $uploadtimeouturl = new moodle_url('/admin/settings.php?section=block_opencast_sharedsettings');
+            $uploadtimeoutlink = html_writer::link($uploadtimeouturl,
+                get_string('uploadtimeout', 'block_opencast'), ['target' => '_blank']);
+            $toolopencastinstanceurl = new moodle_url('admin/settings.php?section=tool_opencast_configuration_' . $instance->id);
```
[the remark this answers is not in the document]
justusdieckmann: I think you didn't address this part

batchupload.php
ferishili: Thanks you for your input,

> You could of course only accept the `series`-param if `$courseid == $SITE->id`, if you'd like that better than checking whether the series is connected to the course.

That check is mandatory, because we are using `$series` also when the are getting back from the `batchupload_form` (submitting the form via add videos button), therefore, if we only accept `$series` in case `$courseid == $SITE->id,` then there is no point of having to check that course-series relation at all!

settings.php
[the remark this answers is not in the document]
ferishili: That is not correct, please check that also in your system and you will find out:

`/admin/settings.php?section=tool_opencast_configuration**_1** `=> points to the (usually default) opencast instance!
`/admin/settings.php?section=tool_opencast_configuration` => throws error!

ferishili:
Hi @justusdieckmann,
I would be very happy to proceed with this PR, therefore, if there is nothing left, I would like to merge this PR.
Thank you for your review.

NinaHerrmann:
@ferishili Please take in mind that @justusdieckmann is not working full-time for us. We are working on Pull Request as fast as we can but we still have an ongoing discussion regarding the batchupload.php file.

ferishili:
> @ferishili Please take in mind that @justusdieckmann is not working full-time for us. We are working on Pull Request as fast as we can but we still have an ongoing discussion regarding the batchupload.php file.

All right, thanks for the info.
I resolved all the requests and answered all questions, please let me know if you have more question.
PS: I admire the strictness on the review, however, I have a feel that is getting a bit more than expected and I don't see any valid objection against this PR!

batchupload.php
justusdieckmann: You're right, although the `$series` we get back from the `batchupload_form` wouldn't need to be checked again, since the mform already makes sure that only the options we pass in can be selected (using server-side validation). If i see it correctly, there is no usecase for redirecting the user to the upload form with a non-$SITE->id course and a $series, so if we rename the url param (for example from series to seriesid), to avoid getting the data from the mform in the `optional_param()`, we could only accept the seriesid url-param if `$courseid == $SITE->id` and otherwise drop the check whether the course is connected to the series.

Of course it also works as it is now, and you don't need to change anything, but you seemed to be unhappy with the current solution, so I just wanted to mention it :)

settings.php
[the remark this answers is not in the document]
justusdieckmann: Hmm, no. I checked in my system and found out:
`localhost/moodle403/admin/settings.php?section=tool_opencast_configuration_1` => throws error!
`localhost/moodle403/admin/settings.php?section=tool_opencast_configuration` => points to my instance.

Did you read that this only happens with only one opencast instance?

ferishili:
> I'm sorry you feel that way. I can see that the two unused imports and unused `global` $variables might not be necessary, but apart from that and the $totalfiles comment, I think that every other reviewer should have mentioned those points I raised. What do you think is more than expected?

Hi @justusdieckmann, no need to be sorry at all.
I think exactly otherwise, I appreciate the review and your time for those unused import and variables as well as string adjustments, but I don't see any useful point or improvement in adding that series checkers which is your only objection against this PR.
Good that you added the [PR for that](https://github.com/Opencast-Moodle/moodle-block_opencast/pull/363), so we can continue there!

All in all, thank you both for your efforts and time, just keep in mind that the main goal is to work together to improve the plugins, not to prove that the other ones work is incorrect!

I would suggest that we schedule a meeting to discuss the point because it appears that our local dev systems are not working the same!

Thanks and best regards

justusdieckmann:
Hey @ferishili,
my only objection against this PR is the link in the settings. As I said, the $series comment in `batchupload.php` was just a completely optional alternative, because I felt that you didn't seem to like the current option. So I'm completely fine with it staying as it is. The pull request I created actually fixes the same thing in `addvideo.php`, since it is currently still a open security vulnerability.

My goal definitely isn't to prove you wrong, but to ensure that we both think that the final result is error-free and well working code.  If we don't do that, a code review doesn't really have a point in my opinion.

On both my machines the link in the settings doesn't work, and I can see that behavior reflected [in the code here](https://github.com/Opencast-Moodle/moodle-tool_opencast/blob/master/classes/settings/admin_settings_builder.php#L78-L86). Please have a look at that, and see if you can tell why it doesn't happen for you.

Thank you very much :)

ferishili:
Hi @justusdieckmann
The tool instance url is covered.
Please check it out.

Thanks and BR

overview_videos.php
```diff
@@ -128,6 +128,14 @@
 $addvideobutton = $OUTPUT->single_button($addvideourl, get_string('addvideo', 'block_opencast'), 'get');
 echo html_writer::div($addvideobutton);
 
+// Show "Add videos (batch)" button.
+if (get_config('block_opencast', 'batchuploadenabled_' . $ocinstanceid)) {
+    $batchuploadurl = new moodle_url('/blocks/opencast/batchupload.php',
+        ['courseid' => $SITE->id, 'ocinstanceid' => $ocinstanceid, 'series' => $series]);
```
NinaHerrmann, line 134: Does this need to be changed to intoseries?

justusdieckmann, line 134: Here the `'series'` also has to be changed to `'intoseries'`

justusdieckmann (approved):
Thank you very much for your work! :)

ferishili:
Thank you very much as well, I am happy that it is done :)

mwuttke:
Hello @taraghi, Hello @llttugraz, if you have some translations for this batch/bulk upload feature you can use [AMOS](https://lang.moodle.org/local/amos/view.php) for the translations into german or you can send me your snippets and I will do that.

And BTW: Many thanks to you for the feature idea and many thanks for the developers and thanks to the developers who realised this! ;-)

mwuttke:
@ferishili, thanks for the exelent work. But I think this feature should be optional. At the moment the batch upload is by default an. Please could you change the admin settings accordingly?

PR #364, opened 2024-03-18
Revert "Make sure teachers can't upload videos into arbitrary series in single-upload"

ferishili:
Reverts Opencast-Moodle/moodle-block_opencast#363


NinaHerrmann:
Why do you think this should be reverted? I won't be available for a meeting today, we can appoint one tomorrow if you think it is not explainable in text.

ferishili:
For the reason I mentioned in that belated review:
The fix has no use to the security issue, because if you pass arbitrary series:
- by query param: it will never reach to the point that you can select it or put it in the form to submit.
- If one can submit anything to the `addvideo` i.e. arbitrary series id: then the provided fix to give null value to `$series` does NOT have any effect!

I can explain it in more details in a live meeting and go through the process step by step

PR #372, opened 2024-03-26
Hidden duplicated videos end up having empty ACLs

### LINKED ISSUE #371: Unexpected behavior by visibility change
Hi,
we noticed unexpected behavior after the visibility change.
When we do an import of hidden video, the video will appear in the LTI series activity, but the video will disappear from the list after few minutes in the Opencast block. All rights on the opencast site are revoked.
After research in task **process_duplicated_event_visibility_change.php**, we came across a condition where the brackets are not there.
https://github.com/Opencast-Moodle/moodle-block_opencast/blob/a2416a1d1aee8536cf6ecba7189acd6083b6c4b3/classes/task/process_duplicated_event_visibility_change.php#L116

With brackets, the video is still displayed and editable on Block Opencast.

Best regards
Alina Antonova

ferishili:
This PR fixes #371 

### Description
Please refer to the issue's description and comments. https://github.com/Opencast-Moodle/moodle-block_opencast/issues/371#issuecomment-2020047954

### Solution
- Added a new argument to the methods: `get_permanent_acl_rules_for_status` and `get_acl_rules_for_status` called `$forceonhidden` which is a flag to return the permanent roles even if the visibility is hidden.
   - The reason to add new argument is to maintain the original concept and to not touch any old behavior.
- if that flag is true then in `get_acl_rules_for_status` in case of hidden we check first if it is permanent and it is forced to return roles, then we prepare the roles and pass it back as the result.

### How to TEST this PR.
- Make sure you are using duplicate event mode for import videos.
- Try to import a mixture of hidden and visible videos in a course to a new one, either by Manual import or Course (bulk) Import wizard.
- Make sure the cronjob is working.
- You would need to wait for both `process_duplicate_event` and ultimately `process_duplicated_event_visibility_change` adhoc tasks to be performed.
  - That means it takes some times for moodle to perform the duplicate event and then when the workflow in opencast is finished the visibility change will take place (so wait until both are done)
- At the end you should confirm that the imported videos (duplicated) have the same original visibility status (except group) in the new course. (hidden->hidden) (visible->visible) (group->hidden)

### DIFF
--- a/classes/local/apibridge.php
+++ b/classes/local/apibridge.php
@@ -1573,12 +1573,13 @@ class apibridge {
      * @param int $courseid id of the course the event belongs to.
      * @param int $visibility visibility of the event.
      * @param array|null $groups array of group ids used for replacing the placeholders
+     * @param bool $forceonhidden flag to force return the acls when hidden.
      * @return array of objects representing acl rules, each with the fields 'allow', 'action' and 'role'.
      * @throws dml_exception
      * @throws coding_exception In case of an invalid visibility status. Only [0,1,2] are allowed.
      */
-    private function get_permanent_acl_rules_for_status($courseid, $visibility, $groups = null) {
-        return $this->get_acl_rules_for_status($courseid, $visibility, true, $groups);
+    private function get_permanent_acl_rules_for_status($courseid, $visibility, $groups = null, $forceonhidden = false) {
+        return $this->get_acl_rules_for_status($courseid, $visibility, true, $groups, $forceonhidden);
     }
 
     /**
@@ -1588,11 +1589,12 @@ class apibridge {
      * @param int $visibility visibility of the event.
      * @param bool $permanent whether to get permanent or non-permanent acl rules.
      * @param array|null $groups array of group ids used for replacing the placeholders
+     * @param bool $forceonhidden flag to force return the acls when hidden.
      * @return array of objects representing acl rules, each with the fields 'allow', 'action' and 'role'.
      * @throws dml_exception
      * @throws coding_exception In case of an invalid visibility status. Only [0,1,2] are allowed.
      */
-    private function get_acl_rules_for_status($courseid, $visibility, $permanent, $groups = null) {
+    private function get_acl_rules_for_status($courseid, $visibility, $permanent, $groups = null, $forceonhidden = false) {
         $roles = $this->getroles($permanent ? 1 : 0);
 
         $result = [];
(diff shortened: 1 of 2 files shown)


classes/local/apibridge.php
```diff
@@ -1614,6 +1616,21 @@ private function get_acl_rules_for_status($courseid, $visibility, $permanent, $g
                 }
                 break;
             case block_opencast_renderer::HIDDEN:
+                if ($permanent && $forceonhidden) {
```
justusdieckmann, line 1619: Sorry, i do not have enough knowledge about ACLs yet; are there no non-permenent hidden ACLs, or why do you require `$permanent == true`?

classes/task/process_duplicated_event_visibility_change.php
```diff
@@ -113,7 +113,7 @@ public function execute() {
             $event = $apibridge->get_already_existing_event([$data->duplicatedeventid]);
             if (!$event || !in_array($event->status, ['EVENTS.EVENTS.STATUS.PROCESSED', 'EVENTS.EVENTS.STATUS.PROCESSING_FAILURE'])
                 || count($event->publication_status) == 0
-                || count($event->publication_status) == 1 && $event->publication_status[0] === 'internal') {
+                || (count($event->publication_status) == 1 && $event->publication_status[0] === 'internal')) {
```
justusdieckmann, line 116: Just to be sure, this is only for readability, and you didn't encounter a bug concerning this, right? Because the behaviour shouldn't be changed by this

justusdieckmann (commented):
Hey @ferishili,

I have added to small questions, thank you very much!

classes/local/apibridge.php
ferishili: Yes, there are non-permanent ACLs (which refer to those that will be used for Students).
And permanent ACLs (refer to those that will be used for Lecturers or Admins)

Permanent ACLs must alway be present in the list no matter the visibility of the event.
However,
NON-Permanent ACLs get removed from the ACL list when the event is hidden.

So here in case of HIDDEN, we only make sure that the permanent one are not empty.

In fact the concept has been implemented to return empty on hidden because in previous stages the permanent ones are already inserted into the ACL list and they won't be changed afterwards. no matter if you toggle the visibility 1000 times.

In this special case, the event is getting duplicated without having the ACLs, therefore the ACL is empty from the beginning! So we need to make that we get permanent ACLs on hidden events as well.

The `$forceonhidden` is only to make sure that we don't accidentally change any behavior of the working code/concept and to use it only for this special case!

Hope that I could explain it clearly!

justusdieckmann: Yes, thank you very much!

PR #373, opened 2024-03-26
Support for new transcription management (subtitles) in Opencast 15

### LINKED ISSUE #368: Subtitles are not available in the Moodle Block Plugin under Manage Transcriptions
If you as a teacher want to handle with existing subtitles in the block plugin for Opencast from Moodle, this unfortunately does not work as expected. The VTT file is available as a direct download. However, it is not available for download under "Action" -> "Manage Transcriptions".

Is it possible that the block plugin cannot yet handle the conversion of subtitles from attachments to tracks?

Our setup:
OS: Debian bookworm
Moodle version: 4.1.9+
Opencast version: 15.3
Moodle block plugin for Opencast Version: v4.3-r2

ferishili:
This PR fixes #368

### Description
In Opencast 15 the subtitles format has been changed drastically, in a form that the flavor now contains captions/delivery and other info like language etc. will be provided in the tags.

### Solution
- In order for us to provide backward compatibility we remain by using `mediapackages` for upload/delete.
- The three cases (attachment, media (old) and media (new)) are covered.
- The manual transcription upload is still following captions/vtt+{lang}, simply because to maintain the difference as well as the lack of capability to upload extra tags info via add_track.

### Goal
the ultimate goal is to be able to manage (delete/download/upload) manual and auto generated subtitles for an event. In order to achieve this goal it might happen that workflows has to be updated (specially `republich-metadata-and-captions` workflow) to cover all cases.

#### NOTE
Consider this PR as a beta version, because this subtitle upgrade is still fresh and might result in unexpected behaviors, therefore, this PR may have further updates


KatrinIhler:
> The manual transcription upload is still following captions/vtt+{lang}, simply because to maintain the difference as well as the lack of capability to upload extra tags info via add_track.

Is this a good idea? We should try and follow the new format. 

In regards to the tagging problem see https://github.com/opencast/opencast/issues/5866

KatrinIhler:
@ferishili Okay, I can see why backwards compatibility would be important. But if this will support the new use case as well (when the tagging PR is through), then that's fine.

justusdieckmann (approved):
Sorry for the delay, the code looks good to me. Thank you very much!

Issue #381, opened 2024-06-14
[block_opencast v4.4-r1] inconsistency between upgrade.php and versions.php


mwuttke:
There is a versionsstring inconsistency between upgrade.php and versions.php

If you try to upgrade the block_opencast plugin from version v4.3-r2 to v4.4-r1 the new table `block_opencast_importmapping` will not be installed, because of the inconsistency between the savepoint number in the [upgrade.php](https://github.com/Opencast-Moodle/moodle-block_opencast/blob/master/db/upgrade.php#L832) and the version number of the [versions.php](https://github.com/Opencast-Moodle/moodle-block_opencast/blob/master/version.php#L28)

We have tested with Moodle 4.1.11 and the block_opencast version 4.4-r1.

NinaHerrmann:
We could reproduce that. The fastest fix would be to adjust the upgrade.php to insert the table conditionally.

mtneug:
@NinaHerrmann @justusdieckmann: can this be closed?

bluetom:
Has been fixed by a merge. See https://github.com/Opencast-Moodle/moodle-block_opencast/pull/382

PR #382, opened 2024-06-14
increased the version number to create table for new version

NinaHerrmann:
currently, it is not possible to upgrade from v4.3-r2 to v4.4-r1
The adjusted upgrade.php adds the table if it does not exist, therefore resolving the issue when releasing a new version while still being compatible with older versions.


NinaHerrmann:
Fixes #381

justusdieckmann (approved):
Looks good!

PR #386, opened 2024-08-08
Failed upload job limiter

### LINKED ISSUE #385: Avoid retrying faulty uploads when upload job process throws error
### Description
It has been determined that, when there is an error during the upload process mostly in opencast side and in this case, when opencast throws "Forbidden" error during the upload to put the video in the assets. Then it results in a bunch of faulty events in opencast without any video but the same metadata. And that is because the cron job does not handle the error properly and processes the faulty job once again in the next iteration.

### Solution
 - Provide a setting to limit the number of iteration after an error occurs: setting it as 0 will result in removing the job directly after the error!
 - Log and inform admin about the situation

ferishili:
This PR fixes #385,

### What is new
- New Setting: in shared setting -> "Limit failed upload attempts"
- Notify teachers if "eventstatusnotificationenabled" is enabled.
- Puts the uploadjob into the “Archived” state when the failed upload limit is reached.
- Teachers are able to:
   - Delete archived uploadjob from the table
   - (NEW) Unarchived Icon "reload icon" is provided right beside the delete icon in the upload job list. This brings back the archived uploadjob to the initial state "Ready to upload" in order to force the cron to pick it up in the next run! aka "unarchive"


ferishili:
There is currently a failure in upstream of `moodle-plugin-ci` which causes the GitHub actions to fail:

https://github.com/Opencast-Moodle/moodle-workflows-opencast/issues/5

tmtsc:
It works!

classes/local/eventstatus_notification_helper.php
```diff
@@ -259,6 +260,48 @@ public static function notify_users_upload_queue($job, $metadata) {
...
+            // Get the role of teachers.
+            $role = $DB->get_record('role', ['shortname' => 'editingteacher']);
+            // Get the course context.
+            $context = context_course::instance($job->courseid);
+            // Get the teachers based on their role in the course context.
+            $teachers = get_role_users($role->id, $context);
+
+            // If teachers array list is not empty, we add them to the user list.
+            if (!empty($teachers)) {
+                foreach ($teachers as $teacher) {
```
justusdieckmann, line 290: This doesn't make sure that the uploader doesn't get notified twice, if they are also a teacher, right?

lang/en/block_opencast.php
```diff
@@ -359,6 +359,8 @@
 $string['expirydate'] = 'Expiry Date';
 $string['failedtogetvideo'] = 'Cannot get event data from Opencast.';
 $string['failedtransferattempts'] = 'Failed transfer attempts: {$a}';
+$string['faileduploadretrylimit'] = 'Limit failed upload attempts';
+$string['faileduploadretrylimitdesc'] = 'Every upload job will be archived after the failed attempts limit is reached. When an upload job is archived, teachers have to decide whether to delete the job or to manually put it back into the queue. <br/> If the event status notification "eventstatusnotificationenabled" is enabled, teachers will get informed about archived status as well. <br/><strong>NOTE:</strong> An empty or 0 (default) value is considered as no limit.';
```
justusdieckmann, line 363: Maybe write that the uploader will always be informed, and the teachers as well, if the setting is enabled

renderer.php
```diff
@@ -625,12 +625,12 @@ public function render_status($statuscode, $countfailed = 0) {
      *
      * @param int $ocinstanceid Opencast instance id.
      * @param array $uploadjobs array of uploadjob objects
-     * @param bool $showdeletebutton shows a delete button in the last column
+     * @param bool $showactionbuttons shows a delete button in the last column
```
justusdieckmann, line 628: The parameter description is outdated

justusdieckmann (commented):
Thank you, this looks great to me! Apart from the notification stuff, I only have one wish/suggestion: [Moodle strongly discourages the use of jQuery in new Code](https://moodledev.io/docs/4.4/guides/javascript/jquery), so if you have the time, it would be great to use native JS instead of jQuery in new Code. But we'll have to go through the javascript code anyway someday and replace all the jQuery, so I would also merge it if you don't do that.

classes/local/eventstatus_notification_helper.php
ferishili: Have you checked this line as well? or did you ask about this line?
https://github.com/Opencast-Moodle/moodle-block_opencast/blob/08d75953571034037854d0bd54e3865ce00803ab/classes/local/eventstatus_notification_helper.php#L292

ferishili:
> Thank you, this looks great to me! Apart from the notification stuff, I only have one wish/suggestion: [Moodle strongly discourages the use of jQuery in new Code](https://moodledev.io/docs/4.4/guides/javascript/jquery), so if you have the time, it would be great to use native JS instead of jQuery in new Code. But we'll have to go through the javascript code anyway someday and replace all the jQuery, so I would also merge it if you don't do that.

Hi @justusdieckmann, thanks for your review.
I have added the changes and answered your question regarding redundant teacher notification.

About the JS and JQuery, I would say, because it is a part of index js file, therefore, it gets a bit messy if I change it to js now, and that would be great, if we could take care of it later on in one single go! Regarding this, I provided new es6 js for my new PR but in here, it is not feasible!

BR

classes/local/eventstatus_notification_helper.php
justusdieckmann: Ok, sorry, I have no idea how in the world I have overlooked that line

ferishili:
Hi @justusdieckmann;
thanks for your review, is there something left to do here, or we could get it merged?

bluetom:
Hi, @ferishili,
I think it's ok for now. I will test it.

PR #398, opened 2024-11-15
Mass Action Support for Video List Tables

### LINKED ISSUE #245: Mass action support for resource entries of courses
Opencast allows to set the visibility of each Video by clicking on the "eye" icon. However, it is not possible to select multiple videos at once.

image

One could consider checkboxes with the action to be chosen from a drop-down menu at the bottom of the table (to be done with mass-actions) in addition to the existing columns.

ferishili:
This PR fixes #245,

## Description
The videos list table in overview page currently does not provide the feature to select all or select some of the videos and perform some generic actions on them. Refer to issues comments for more info!

## How it works:
- It targets two major video tables in index page and overview_video page.
- On the table header there is a checkbox now that select/deselect all videos
- Each table row now gets a very first column containing a checkbox to select/deselect that video.
- At the top and bottom of the table now there is a dropdown selection with the label "With Selected Videos..." that contains the following actions:
   - Delete
   - Update metadata
   - Change Visibility
   - Start workflow
   NOTE: These actions exist already and could be performed also for each video from the action menu dropdown!
- By default at the first page load:
   - no video is selected.
   - the mass action dropdown is disabled, and will be enabled only when video(s) are selected via the checkboxes.
- When a confirmation dialogue is closed, the video (Mass-Action) selection will go back initial state (no selection).

### Mass Actions:
- A confirmation dialogue will be displayed when actions are selected via the dropdown, except the "**Start workflow**", this action will display the Start Workflow modal but with confirmational texts. The reason behind this behavior is to avoid opening modal with modal with is a bad UX/UI way to do!
- For **Update metadata** in mass action, it is important to provide the "Enable" buttons in the metadata form, in order to enable and edit a metadata field for multi videos intentially, not to clear the rest if you only want for example change subject, and end up removing description!
- Important to for the success/error and info notifications which will be notified after an action is performed:
  - In order to avoid complexity in text and meanings, there will be separate notifications for the actions in terms of success/error and info to report the end result.
- Providing an action depends on the permissions as well as the configurations if any!

## Testing:
- Patch this PR
- Have a course with multiple functional videos.
- Go to the overview page of this plugin.
- Target the mass action additions in the first column of the rows as well as top and botton area of the table with dropdowns "With Selected Videos...".
- Test the selection behavior of the mass actions:
   - No selection and disabled dropdown at first.
   - Select all checkbox must select/deselect all checkboxes and enable the dropdown.
   - If on video is also selected while others are not, the select all checkbox at the header must be deselected, but the dropdown must become enabled.
- Based on the configuration and permissions, the mass actions must be provided:
   - Delete: `block/opencast:deleteevent` permission
   - Update metadata: `block/opencast:addvideo` permission
   - Change Visibility: `block/opencast:addvideo` permisison + `workflow_roles` & `aclcontrolafter`configurations.
   - Start Workflow: `block/opencast:startworkflow` permission + `workflow_tags`configuration
- Test all provided actions by selection them from the dropdown:
   1. the modal confirmations + "Start workflow" enhanced modal
   2. The editing page in case of Change Visibility and Update metadata
   3. Pay attention to the notifications (alerts) of the actions, when they are finished!


mwuttke:
In my test environment, based on the[ lms-community-integration](https://github.com/opencast/community-integrations/tree/main/lms), the PR works as expected.

bluetom (approved):

PR #400, opened 2024-11-20
Applying correct default completion while building LTI module

### LINKED ISSUE #377: Providing single video as an lti module doesn´t work
Hi there,
we have following installation:
Opencast 15.1
Moodle 4.1
block_opencast v4.3-r2
tool_opencast v4.3-r1

Problem:
When providing a single video as an activity for students to watch directly in Moodle, I get following error in Moodle.

Bildschirmfoto 2024-04-30 um 09 29 17

We do not have opencast_mod installed.

But as a test we installed opencast_mod (v4.3-r1) and deactivated it in Moodle and the providing of the single Video worked fine.

This issue is somehow related to: https://github.com/Opencast-Moodle/moodle-block_opencast/pull/361

Greetings Raffaela

ferishili:
This PR fixes #377 

## Description
The problem occurs when admin decides to use LTI module in `block_opencast` only without installing Opencast Activity Plugin `mod_opencast`. When creating an LTI module via `block_opencast`, it ended up throwing error that complains about missing opencast activity module, which forces to have that plugin!

## Solution
This fix applies the correct/related module name "lti" and set the must exists flag, instead of "opencast" module or the `mod_opencast` plugin.
By adding `MUST_EXIST`, we make sure that lti module is a requirement to create lti module in opencast which is the right thing to do!

## Testing
- Patch this PR.
- Make sure to have LTI preconfigured external tool ready.
- Make sure to correctly configure the LTI modules in admin settings of the this plugin!
- Make sure to **not** to have the `mod_opencast` plugin installed.
- Have a course with the opencast block that has at least one video ready.
- Go to the video list and try to make the LTI module (LTI Provider) for the video.
- Result **with** this patch would be to: successfully creating LTI module provider.
- Result **without** this patch would be to see the error `invalid plugin_support mod_`

### DIFF
--- a/classes/local/ltimodulemanager.php
+++ b/classes/local/ltimodulemanager.php
@@ -443,7 +443,7 @@ class ltimodulemanager {
         $moduleinfo->availability = $availability;
 
         // Apply completion defaults.
-        $module = $DB->get_record('modules', ['name' => 'opencast']);
+        $module = $DB->get_record('modules', ['name' => 'lti'], '*', MUST_EXIST);
         $defaults = manager::get_default_completion($course, $module);
         if ($module) {
             foreach ($defaults as $key => $value) {


bluetom:
I think it's ok, but just for my understanding: That means, you get the default conditions of a(ny) LTI module?

ferishili:
> I think it's ok, but just for my understanding: That means, you get the default conditions of a(ny) LTI module?

not just any, but the one specifically configured for the course which happens in `get_default_completion` method.

bluetom (approved):

bluetom:
We in Münster do not have the LTI-modul in use anymore and I have not been successfull so far to get it runnung in my test environment. So I accept this PR because of the successfull checks and the code review and there seems to be nothing which could cause big problems.

PR #405, opened 2024-12-16
Use new opencast api exceptions from API plugin

ferishili:
This PR is to apply the changes introduced in https://github.com/Opencast-Moodle/moodle-tool_opencast/pull/70

## NOTE:
The `branch-tool-plugin` in the `moodle-ci.yml` file has to be reverted to `main` after merging this PR!


bluetom (commented):
Does this PR/code really only work with opencast version higher than 16.6?

